PR #8, opened 2018-04-12
Add import 3 libraries for new test case

SyHuangCBN:
1. pysnmp: It's for modem SNMP communication
2. pyasn1: It's for modem SNMP communication
3. selenium: It's for getting data from web site

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ ipaddress
 pysnmp
 selenium
 pyasn1
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,6 @@ dlipower
 cdrouter
 pyserial
 ipaddress
+pysnmp
+selenium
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -10,3 +10,6 @@ dlipower
 cdrouter
 pyserial
 ipaddress
+pysnmp
+selenium
+pyasn1
```
mattsm: Can we add these in your own layer? We will make our automation scripts run pip install on all the overlays if the file exists.

boardfarm-cbn/requirements.txt

requirements.txt
```diff
@@ -13,3 +13,4 @@ ipaddress
 pysnmp
 selenium
 pyasn1
+
```
mattsm: This is an empty pointless commit as far as I can tell

PR #20, opened 2018-04-26
Update Cyber Power switch for 24 ports

SyHuangCBN:
Cyber power switch support telnet and SNMP protocal

This case used telnet to control each port on/off


devices/power.py
```diff
@@ -296,10 +296,93 @@ def off(self):
...
+        pcon.send('4' + "\r\n")
+        pcon.expect("Goodbye")
+
     def reset(self):
-	self.off()
-	time.sleep(5)
-	self.on()
+        self.off()
+        time.sleep(5)
+        self.on()
```
mattsm: You took this from the above class so it is missing a reset() function now. We need to add it for your class and leave it on the other class.

mattsm:
I fixed the one comment for you, and pushed these changes. Closing pull request since the changes are merged.

PR #21, opened 2018-04-26
devices: power: if decode fails, try to use raw data

mattsm:
It's possible the data we have might still work, so let's proceed
without the decode


mattsm:
The automated test failed, can you add jira module to requirements.txt as well?

PR #26, opened 2018-05-09
devices: debian: add wan-no-eth0 option

mattsm:
It's possible to have a container that has no eth0 and eth1 is the only
connected interface to the board and outside. So let's add an option to
skip using that interface

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -26,6 +26,7 @@ class DebianBox(base.BaseDevice):
     prompt = ['root\\@.*:.*#', '/ # ', ".*:~ #" ]
     static_route = None
     wan_dhcp = False
+    wan_no_eth0 = False
     wan_cmts_provisioner = False
 
     def __init__(self,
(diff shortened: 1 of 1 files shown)


SyHuangCBN:
It doesn't have problem.
Tested by SY.

SyHuangCBN:
We tried to run command, but it always run another command "create_container_eth1_vlan $vlan"
./deploy-boardfarm-nodes.sh create_container_eth1_static 172.16.1.154 172.16.1.1

How to run command correctly to run " create_container_eth1_static " ?

Because we always run "create_container_eth1_vlan $vlan" command

echo "Creating nodes starting on vlan $START_VLAN to $END_VLAN on iface $IFACE"

for vlan in $(seq $START_VLAN $END_VLAN); do
	echo "Creating node on vlan $vlan"

	**create_container_eth1_vlan $vlan**

	[ "$OPTS" = "both" ] && { local_route; continue; }
	if [ $((vlan%2)) -eq 0 ]; then
		[ "$OPTS" = "even" ] && local_route
	elif [ "$OPTS" = "odd" ]; then
		local_route
	fi
done

PR #27, opened 2018-05-09
Pr fix local route for containers

mattsm:


mattsm:
You need to drop the ** - that's markdown syntax. Not sure why it did not show as bold:

set -x; IFACE=eth1 create_container_eth1_static wan-shared 172.16.1.154/24 172.16.1.1; set +x

PR #28, opened 2018-05-14
Pr flash meta after boot

mattsm:


SyHuangCBN:
Tested by SY.
It's doesn't have issue.

PR #30, opened 2018-05-16
Pr tftp ipv6 and start syslog

mattsm:


SyHuangCBN:
When we use command on local boradfarm to test, but still have the same problem.
It's can't connect 172.16.1.154. I known what you maen it can't allow to test on local machine.

But When we use ssh to test boardfarm. We have another problem below.
https://github.com/lgicbn/boardfarm-cbn/pull/27

PR #32, opened 2018-05-24
devices: debian: remove duplicate time-servers option in cmts dhcp co…

mattsm:
…nfig

This was causing the wrong time server to be set on the server

Reported-by: Rachel Lei <rlei@libertyglobal.com>


SyHuangCBN:
Tested by SY.
It doesn't have issue.

PR #36, opened 2018-05-29
support ipv6 dhcp

rachellei:


devices/debian.py
```diff
@@ -351,11 +351,21 @@ def setup_as_cmts_provisioner(self, board_config):
...
         self.sendline('ip route add 192.168.201.0/24 via 192.168.3.222')
         self.expect(self.prompt)
         self.sendline('ip route add 192.168.200.0/24 via 192.168.3.222')
         self.expect(self.prompt)
+        self.sendline('ip -6 route add 2001:ed8:77b5:2000::/64 via 2001:ed8:77b5:3::222 dev eth1  metric 1024')
+        self.expect(self.prompt)
+        self.sendline('ip -6 route add 2001:ed8:77b5:2001::/64 via 2001:ed8:77b5:3::222 dev eth1  metric 1024')
+        self.expect(self.prompt)
+        #check if 'dadfailed' happen  
+        self.sendline('ip addr show eth1')
```
mattsm: You should really check the output with expect and look for the 'dadfailed' - at least print a warning if we see it but should probably fail the test. It seems like we are broken in this state.

devices/debian.py
```diff
@@ -418,6 +511,10 @@ def setup_as_cmts_provisioner(self, board_config):
         self.sendline('/etc/init.d/isc-dhcp-server start')
         self.expect(['Starting ISC DHCP(v4)? server.*dhcpd.', 'Starting isc-dhcp-server.*'])
         self.expect(self.prompt)
+        self.sendline('touch /var/lib/dhcp/dhcpd6.leases')
+        self.expect(self.prompt)
+        self.sendline('/usr/sbin/dhcpd -f -6 -cf /etc/dhcp/dhcpd6.conf -pf /var/run/dhcpd6.pid &')
+        self.expect(self.prompt)
```
mattsm: We can't setup the configuration such that it starts with '/etc/init.d/isc-dhcp-server start' command above?

devices/debian.py
rachellei: I did see the dadfailed if I add ipv6 address to eth1 manually which we don't know the reason yet. But with adding the ipv6 address via boardfarm code I don't see it happens during many times of testing.

mattsm: You can add a 
```
if 0 == self.expect(['dadfailed'] + prompt):
    raise Exception("ipv6 dadfailed present, ipv6 will not funtion")
```

devices/debian.py
rachellei: I did try this yesterday. I looked into the file isc-dhcp-server. I'm not sure if modifying isc-dhcp-server is better solution?

mattsm: It's probably just something mis-configured? It would be nice, we do run the restart in a few places to reload configs for example.

mattsm: Correction: I think we need to upgrade our debian nodes and ipv6 should have better support via the initscripts.

mattsm:
Closed in favor of reworked version https://github.com/lgirdk/boardfarm/pull/38

PR #39, opened 2018-06-14
devices: casa_cmts: update to use connection decider

mattsm:
This waits properly for various connection methods to be established
before attempting to run commands


SyHuangCBN:
Tested by SY.
It doesn't have issue.

PR #40, opened 2018-06-14
devices: debian: bypass DAD errors for now

mattsm:
We are using 4 gige link to our switch for our devices. We will get DAD
failures (dupliate address detection) if the switch is not configured to
do LAG properly.

We need to revist the switch config to fix this issue but to unblock
ipv6 we do this for now


SyHuangCBN:
Tested by SY.
It doesn't have issue.

PR #47, opened 2018-06-20
dual-stack to support ipv4 and ipv6 at the same time

rachellei:
Feature: both ipv4 and ipv6 work at the same time.
What's needed with these request together to support the above feature.
1. CMTS config: interface docsis-mac 1 ip-provisioning-mode dual-stack
2. boardfarm-cbn devices/mv1.py add 9_EU_CBN_Dual-Stack_LG.cfg to config['tftp_cfg_files'].
3. "flags = IPv6" -> /etc/xinetd.d/time which Matthew will add it.


mattsm:
Coming soon, but not via this commit...

PR #53, opened 2018-06-28
devices: power: add IP9820 power switch section

SyHuangCBN:
Due to we still have IP9820 power swith to use.
We need to add power switch section.

Signed-off-by: sy_huang <sy_huang@compalbn.com>

### DIFF
--- a/devices/power.py
+++ b/devices/power.py
@@ -64,6 +64,8 @@ def get_power_device(ip_address, username=None, password=None, outlet=None):
         return Ip9258(ip_address, outlet, username=username, password=password)
     if 'Cyber Power Systems' in data:
         return CyberPowerPdu(ip_address, outlet=outlet, username=username, password=password)
+    if 'Unauthorized' in data:
+        return Ip9820(ip_address, outlet)
     else:
         raise Exception("No code written to handle power device found at %s" % ip_address)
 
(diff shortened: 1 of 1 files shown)


SyHuangCBN:
I saw it shows failure, not sure what's failure with IP9820. Will keep tracking.

devices/power.py
```diff
@@ -64,6 +64,8 @@ def get_power_device(ip_address, username=None, password=None, outlet=None):
         return Ip9258(ip_address, outlet, username=username, password=password)
     if 'Cyber Power Systems' in data:
         return CyberPowerPdu(ip_address, outlet=outlet, username=username, password=password)
+    if 'Unauthorized' in data:
+        return Ip9820(ip_address, outlet)
```
mattsm: Why Unauthorized? That might match other http data in other devices. Can you please share the full output?

devices/power.py
```diff
@@ -330,6 +332,45 @@ def reset(self):
...
+        return urllib2.urlopen('http://' + self._ip_address + '/set.cmd?cmd=setpower+p6' + str(self.port) + '=0')
+
+    def reset(self):
+        self.off()
+        time.sleep(5)
+        self.on()
+
+    def getPowerStatus(self):
+
+        oid = self.oid_Outlet + '.' + str(self.port) + '.0'
```
mattsm: Where does self.oid_Outlet get set? Please just remove these functions if they were used for testing.

devices/power.py
lynnlincbn: We can't get power device name to mask sure which power device in use.
Only can get below message, so we use 'Unauthorized' to check power device is Ip9820.

Please refer below http data message:

<html><head><title>Document Error: Unauthorized</title></head>
--
  | <body><h2>Access Error: Unauthorized</h2>
  | <p>Access to this document requires a User ID</p></body></html>

devices/power.py
lynnlincbn: Will remove it, thanks.

devices/power.py
mattsm: I understand that, but I think you need to try to login and see what model this is... if you just do Unauthorized you will match ANY device that has that text in their output. Well that's more than just this device. You need to think generically.

I would say, if you don't match anything without logging in- then you do just that login and see if you can find a model number in the output and match off that.

mattsm:
Also, please rebase this change on the latest master... we've since fixed the travis ci issues.

SyHuangCBN:
Update #57 
Closed it.

PR #57, opened 2018-07-02
Pr add ip9820 power switch v2

SyHuangCBN:
1. Add IP9820 power switch section
2. Add detect firefox function
3. Remove IP9258 power switch section
4. Add IP9280 class section

### DIFF
--- a/devices/power.py
+++ b/devices/power.py
@@ -295,25 +295,6 @@ class SimpleSerialPower(PowerDevice):
 
             ser.close()
 
-#
-# IP Power 9258 networked power switch class
-#
-# This work is released under the Creative Commons Zero (CC0) license.
-# See http://creativecommons.org/publicdomain/zero/1.0/
-
-# Example use:
-#
-# import time
-# from ip9258 import Ip9258
-#
-# ip9258 = Ip9258('192.168.1.10', 'admin', 'password')
-#
-# for i in range(4):
-#     ip9258.on(i)
-#     time.delay(1)
-#
-#     ip9258.off(i)
-#     time.delay(1)
 
 import urllib2
 
--- a/devices/power.py
+++ b/devices/power.py
@@ -81,6 +81,31 @@ def get_power_device(ip_address, username=None, password=None, outlet=None):
     else:
         raise Exception("No code written to handle power device found at %s" % ip_address)
 
+def detectFirefox(dirpath='/usr/bin'):
+
+    """
+    Name: detectFirefox
+    Purpose: detect firefox on platform before run test
+    Input: firefox dirpath on linux
+    Output: nul
+    """
+
+    driver_url_linux32='https://github.com/mozilla/geckodriver/releases/download/v0.20.1/geckodriver-v0.20.1-linux32.tar.gz'
+    OS=platform.system()
+
+    # detect firefox existed on Linux
+    if OS=='Linux':
+        firefox_isexist=os.popen('which firefox').read().strip('\n')
+        pattern = re.compile(r'Mozilla Firefox (\d+)..*')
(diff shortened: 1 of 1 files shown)


devices/power.py
```diff
@@ -50,7 +51,18 @@ def get_power_device(ip_address, username=None, password=None, outlet=None):
...
+        """ Firefox """
+        options = webdriver.FirefoxOptions()
+        options.set_headless()
+        options.add_argument('--disable-gpu')
+        driver = webdriver.Firefox(firefox_options=options)
+        driver.get("http://" + username + ':' + password + '@' + ip_address)
+        data = driver.page_source
+        driver.find_element_by_id('memLogout').click()
+        time.sleep(1)
+        driver.quit()
```
mattsm: No, this is not what I meant.

If there were not match just login with the previous method you were using and check the output after logging in. DO NOT START FIREFOX FOR THIS.

devices/power.py
```diff
@@ -81,6 +81,31 @@ def get_power_device(ip_address, username=None, password=None, outlet=None):
...
+        pattern = re.compile(r'Mozilla Firefox (\d+)..*')
+        match = pattern.match(os.popen('firefox -v').read())
+        firefox_v=match.group(1)
+        assert (firefox_isexist!='' and int(firefox_v)>=60), 'not found: please install firefox 60+'
+
+        driver_path=File_findPath('geckodriver', path=dirpath)
+        if driver_path is '':
+            os.system('sudo rm %s/geckodriver*.tar.gz' % dirpath)
+            os.system('sudo wget -P %s/ %s' %(dirpath, driver_url_linux32))
+            os.system('sudo tar zxvf %s/geckodriver*.tar.gz -C %s/' %(dirpath, dirpath))
```
mattsm: Boardfarm scripts will never ever detect firefox anywhere. EVER. Please stop trying to add this function.

-M

devices/power.py
```diff
@@ -295,25 +295,6 @@ def reset(self):
...
-# from ip9258 import Ip9258
-#
-# ip9258 = Ip9258('192.168.1.10', 'admin', 'password')
-#
-# for i in range(4):
-#     ip9258.on(i)
-#     time.delay(1)
-#
-#     ip9258.off(i)
-#     time.delay(1)
```
mattsm: Don't remove this, it was working right? It can stay for now for other users that might have this power switch type.

devices/power.py
```diff
@@ -350,6 +350,35 @@ def reset(self):
...
+        self._ip_address = ip_address
+        self.port = port
+
+        # create a password manager
+        password_mgr = urllib2.HTTPPasswordMgrWithDefaultRealm()
+        password_mgr.add_password(None, 'http://' + ip_address, username, password)
+        handler = urllib2.HTTPBasicAuthHandler(password_mgr)
+        opener = urllib2.build_opener(handler)
+        # Now all calls to urllib2.urlopen use our opener.
+        urllib2.install_opener(opener)
```
mattsm: This is how you can check the output of the main page if it requires a login... don't use firefox.

devices/power.py
lynnlincbn: Will remove it.

devices/power.py
lynnlincbn: Will remove it.

devices/power.py
lynnlincbn: Will restore it.

devices/power.py
lynnlincbn: Will update pull request to remove use firefox function.
I can get power device name by you provided method.
Thanks for your help.

SyHuangCBN:
Update #61 for code review.
Closed it.

PR #60, opened 2018-07-03
Pr misc error msg etc

mattsm:


mattsm:
Merging first two manually, waiting on the install pkg consolidation

PR #62, opened 2018-07-06
devices: base.py:fix_get_interface_ip6addr

lynnlincbn:
Fix regexp to get ipv6 function

ex:
inet6 addr: fe80::40ec:26b8:95a5:1132/64
function return 'fe80::40ec:26b8:95a5:1132/64' change to 'fe80::40ec:26b8:95a5:1132'

Signed-off-by: luke_tseng <luke_tseng@compalbn.com>


mattsm:
I think at some point we will want to return both values... but merging for now.

PR #64, opened 2018-07-09
devices: debian: more fixups for installing packages

mattsm:


mattsm:
I'm merging this now because it's failing on CBN lab network because their internet is not working on all machines.

PR #68, opened 2018-07-18
devices: windows_telnet: add inital device for a windows telnet device

mattsm:
This connects to a windows machine over telnet. Here is a basic example
test:
```
from devices import winwifi

class wintest(rootfs_boot.RootFSBootTest):
    def runTest(self):
        winwifi.sendline('dir')
        winwifi.expect(winwifi.prompt)
```
You can add a device in the boarfarm config via:
```
"devices": [
		{
			"name": "winwifi",
			"type": "windows-telnet",
			"ipaddr": "$ipaddr",
			"username": "user",
			"password": "pass"
		}
```

### DIFF
--- a/devices/windows_telnet.py
+++ b/devices/windows_telnet.py
@@ -1,0 +1,33 @@
+
+import sys
+import base
+import connection_decider
+
+class WindowsTelnet(base.BaseDevice):
+
+    model = ('windows-telnet')
+    # This prompt regex could use more work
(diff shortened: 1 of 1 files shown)


lynnlincbn:
I can run voice test without call windows pc to run script now.
So we will keep your suggest, if we run tcl program need to call windows pc will used your method.
Thanks for your help.

PR #71, opened 2018-07-26
Request for add the required python components

luketseng:
tftpy: it's needed for copy file to cnr server by tftp.
shutil: it's needed for file operating. ex: copy, move...

Signed-off-by: luke_tseng <luke_tseng@compalbn.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,5 @@ jira
 simplejson
 pyvirtualdisplay
 Faker
+tftpy
+shutil


requirements.txt
```diff
@@ -15,3 +15,5 @@ jira
 simplejson
 pyvirtualdisplay
 Faker
+tftpy
+shutil
```
mattsm: These are not used in this repo. We will not add them.

mattsm: Also shutil is not something you pip install.

luketseng: got it, thx.

PR #80, opened 2018-08-21
devices: bft: modify result and log path

NeroChenCBN:
add result rename and create path .
Path will be result/"CM Name, EX. mv1-2-1"/date/date_time_result.htmlwq

Signed-off-by: NeroChenCBN <nero_chen@compalbn.com>

### DIFF
--- a/bft
+++ b/bft
@@ -443,11 +443,13 @@ def main():
 
     # Create Pretty HTML output
     import make_human_readable
+    output_name=config.output_dir+config.BOARD_NAMES[0]+"/"+time.strftime("%Y%m%d")+"/"
     try:
         title_str = make_human_readable.get_title()
         make_human_readable.xmlresults_to_html(full_results['test_results'], title=title_str,
-                                output_name=os.path.join(config.output_dir, "results.html"),
+                                output_name=output_name,
                                 board_info=config.board)
+
     except Exception as e:
         print(e)
         print("Unable to create HTML results")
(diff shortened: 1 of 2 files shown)


bft
```diff
@@ -443,11 +443,13 @@ def main():
 
     # Create Pretty HTML output
     import make_human_readable
+    output_name=config.output_dir+config.BOARD_NAMES[0]+"/"+time.strftime("%Y%m%d")+"/"
```
mattsm: Other things put things in config.output dir - isn't it easier to just update this?

I think we want to keep the behavior the same, so maybe adding a symlink to a new dir is the easiest approach?

NeroChenCBN: We can update config.output_dir, but I cannot find where should I modify. Can you tall me?
I think that is for our use  , or do you have any other consider?

mattsm: Does this do what you need?

[mattsm@goliath boardfarm ((7af5c47...))]$ ./bft --help | grep output_dir
  -o , --output_dir     Directory to output results files too

NeroChenCBN: Not this .
We need this because there have Jinkens trigger and manual trigger, and we cannot control manual what time trigger it.
If use -o is too trouble for manual testers and the log name will same with previous.

We hope automatically and clearly distinguish each execution result like the code i was commit.
It can auto distinguish  by date , time and cm name ,pull log and html to corresponding folder and add date time to result name.

EX.  
when a test run down in 20180907 10:20:33 AM , the log path is follow:
  result/MV1-1-14/20180907/20180907-102033_result.html
  result/MV1-1-14/20180907/NTL7465LG__P1__85-20180907-102033.txt

mattsm: You can just run:

./bft .... -o results/mv1-$(date +%m%d%H%M%y%s)

The station name is in the results in that directory.

-M

mattsm: if you make a pull request where you don't break every other user that replies on the output dir being 'results' then I will consider your pull request. that is not what you have here.

mattsm:
This is old, place reopen if you want to discuss more.

PR #83, opened 2018-09-04
Pr add support for feature flag

mattsm:


lynnlincbn:
Tested by Lynn.
It doesn't have issue and I will commit cbn.json to set "true" of "available_for_autotests".

PR #84, opened 2018-09-05
devices: debian: make static route for lan device as well

mattsm:
Keep static route for wan with the 'wan-' prefix and make this work for
both wan and lan if required.

This let's us keep static routes set for wan and lan devices. The use
case here is to keep the main SSH connection established to the device.


mattsm:
@SyHuangCBN any issues?

@mgualco-contractor as well

PR #85, opened 2018-09-06
deploy-boardfarm-nodes: ping default route

mattsm:
Wake up these containers so the upstream switch knows about these mac addresses


mattsm:
@mgualco-contractor

PR #86, opened 2018-09-06
devices: local_cmd: skip to next device if cmd closes

mattsm:
This will simply fail if we don't connect to the device via this
comamnd. Let's raise an exception instead so it backs out and trys
another board if one is available


mattsm:
@mgualco-contractor

PR #95, opened 2018-09-19
deploy-boardfarm-nodes: pick docker0 as interface to route via

mattsm:
Previous commit changed this from the default docker route, so let's just
ask the interface instead


luketseng:
I test this commit ok.
It can merge to master, thx.

PR #98, opened 2018-09-27
Pr more fixes for debian as dynamic device

mattsm:


mgualco-contractor:
I tested it and it works now, thx

PR #113, opened 2018-10-11
devices: base.py: modify regex of get_interface_ip6addr()

luketseng:
- inet6 format is different between board.get_interface_ip6addr("wan0") and wan.get_interface_ip6addr("eth1")
    - 'inet6 addr: 2001:b030:701d:2100:b593:1b01:313:ae00/128 Scope:Global' for board.get_interface_ip6addr("wan0")
    - 'inet6 2002:db50:fa13:1:78fe:23ff:feef:f0ac  prefixlen 64' for wan.get_interface_ip6addr("eth1")
    - modify regex to self.expect('\s((200([0-9a-f]){1,1}:)([0-9a-f]{1,4}:){1,6}([0-9a-f]{1,4}))([\/\s])', timeout=5)

### DIFF
--- a/devices/base.py
+++ b/devices/base.py
@@ -31,7 +31,7 @@ class BaseDevice(pexpect.spawn):
 
     def get_interface_ip6addr(self, interface):
         self.sendline("\nifconfig %s" % interface)
-        self.expect('inet6 addr: (200(.+))/\d+ Scope:Global', timeout=5)
+        self.expect('\s((200([0-9a-f]){1,1}:)([0-9a-f]{1,4}:){1,6}([0-9a-f]{1,4}))([\/\s])', timeout=5)
         ipaddr = self.match.group(1)
         self.expect(self.prompt)
         return ipaddr


luketseng:
> I think we need to drop the 200 prefix as well? I will leave it for now since it's there but what was the reasoning behind that?

The reason for this is as below if sendline.('ifconfig') and get inet6.
You might get inet6 with fe8 prefix if drop the 200 prefix.

root@bft-node-eth1-wan-shared:~# ifconfig
eth1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.16.1.247  netmask 255.255.255.0  broadcast 0.0.0.0
        **inet6 fe80::78fe:23ff:feef:f0ac  prefixlen 64  scopeid 0x20<link>**
        **inet6 2002:db50:fa13:1:78fe:23ff:feef:f0ac  prefixlen 64  scopeid 0x0<global>**
        ether 7a:fe:23:ef:f0:ac  txqueuelen 1000  (Ethernet)
        RX packets 2933718  bytes 625248844 (596.2 MiB)
        RX errors 0  dropped 162033  overruns 0  frame 0
        TX packets 11340692  bytes 16234857032 (15.1 GiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

mattsm:
Yea, we should possibly return a list, and filter link addresses. This is OK for now but once we make more changes let's give it some serious thought.

PR #124, opened 2018-10-17
tests: lib: common.py: add firefox options fix error

luketseng:
- error Message: Process unexpectedly closed with status 1
- add firefox options: opts.add_argument("--headless") to fix it

Signed-off-by: luke <luke_tseng@compalbn.com>

PR #129, opened 2018-10-18
tests: screenshot_gui: save display in self.display

mattsm:
So we can run recover properly


mattsm:
@luketseng please review

luketseng:
It is ok. But I run ScreenshotGUI test case always fail.
https://github.com/lgirdk/boardfarm/pull/124

PR #138, opened 2018-11-05
Casa cmts is cm bridged

mgualco-contractor:
helper function that returns True if the CM is in bridge mode


mgualco-contractor:
Merged (cherry picked, leaving out the two merge commits)

PR #141, opened 2018-11-07
devices: base.py: add .strip() to return string

luketseng:
- self.before.strip() instead of self.before


luketseng:
Have any problem?

luketseng:
I ask to slack.
It seems that no one will affect.

PR #146, opened 2018-11-12
deploy-boardfarm-nodes: add static ip option with docker network intact

mattsm:


luketseng:
Good. It's working on BF1.
I will change BF1~BF2 server environment(172.16.10.x network segment) and change cbn.json after this commit merge.

PR #152, opened 2018-11-13
deploy-boardfarm-nodes: fix random mac function

mattsm:
$ . deploy-boardfarm-nodes.sh
$  random_private_mac
00:16:3e:30:83:f0

### DIFF
--- a/deploy-boardfarm-nodes.sh
+++ b/deploy-boardfarm-nodes.sh
@@ -8,7 +8,18 @@ BRINT=br-bft
 BF_IMG=${BF_IMG:-"bft:node"}
 
 random_private_mac () {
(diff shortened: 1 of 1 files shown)


mattsm:
This was meant to spark discussion.. this address range seems to be some common one delegated to VMs so I started here. I will make the changes you requested after reviewing them.

mattsm:
Just to confirm, we are looking at these:

x2:xx:xx:xx:xx:xx
x6:xx:xx:xx:xx:xx
xA:xx:xx:xx:xx:xx
xE:xx:xx:xx:xx:xx

That is private and non-multicast?

preddy-contractor:
Now we can ignore this 7th bit to create MAC-Address since it uses only to create EUI-64 address for IPv6 interface.
Definition: 7th bit uses to create the EUI-64 address, this address can be created either Link-local or global address for IPv6 interface. 7th bit toggles or inverts to do this operation from mac-address.
Example:
Mac-Address: fa16.3e60.0218
1) fa = 1111 1010 ---> f8 = 1111 1000
2) f8 = 1111 1000 ---> fa = 1111 1010

PR #163, opened 2018-11-18
tests: lib: regexlib.py regular expressions library

mgualco-contractor:
this should become a one stop shop for common regex


mattsm:
Thanks for this, can you make some addl simple replacements for already existing code?

PR #165, opened 2018-11-19
tests/lib/docsis.py: pr_add_docsis_md5_checK_function

luketseng:
- md5sum_check(encode_cfg_path_from_tmp_dir)
    return True if the content of cfg is same as original file.
- you can use as below to check file
    assert self.md5sum_check(cmcfg_path), 'md5sum_check() fail'

### DIFF
--- a/tests/lib/docsis.py
+++ b/tests/lib/docsis.py
@@ -19,6 +19,8 @@ class docsis:
             return output file name(.txt)
         encode(output_type='cm_cfg')
             return output file name(.cfg or .bin)
+        md5sum_check(encode_cfg_path_from_tmp_dir)
+            return True if the content of cfg is same as original file.
     """
     def __init__(self, file_path):
         self.file_path=file_path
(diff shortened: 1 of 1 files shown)


mattsm:
What will this be used for? The purpose is not clear from the commit message.

SyHuangCBN:
Please refer link below.
https://jira.compalbn.com/browse/AUTO7465LG-15

mattsm:
@SyHuangCBN please reply with the comments again, don't post private links please.

luketseng:
This commit can check encode of cm_cfg txt is correct.
step is 
1. copy cm_cfg .txt file to 'tmp' dir.
2. encode original .txt to .cfg
3. decode .cfg to txt
4. check txt is same as original .txt and check done

mattsm:
Please don't use all these os.system() and use python libraries directly...

@mgualco-contractor we can use this to check if the CM cfg changed... or something like this...

luketseng:
Have any questions?

mattsm:
This is not really an md5 check either, you are using the md5 signature of the files to see if they are producing equivalent CM cfgs. This is a useful thing to have I suppose (esp. given the issues we've seen with encoding MTA cfgs).

However, let's call it verify_encode_decode_equivalence() - fixup the name and try to use builtin python functions for md5 sum and equivalent sed commands although that would not be the end of the world.

And please fix-up the commit message to be more clear what this new function is doing.. md5_check() sounds like it's verify the file integrity and not verify the docsis tool is working as expected.

luketseng:
Can we give up this commit if we think that it's not really an md5 check?
For now, we want to use md5 tool check content is same between original txt and decode txt.
But decode txt will output rand code as below and lead to we need modify content before use md5.
So, Can we update this issue after we find best way to chech?

mattsm:
OK, I think this is a decent way.. but also just running difftool on the two files after removing the offending lines could work as well.

Anyways I was just asking to rename the function to what the intention of the function is supposed to be.

PR #176, opened 2018-11-29
Pr new wifi tests

sravichandran-contractor:
SNMP function for setting SSID and Password.
Connect to wifi and ping to the internet
SNMP mib parameter functions for WIFI


tests/ssid_set_snmp.py
```diff
@@ -0,0 +1,40 @@
...
+# All rights reserved.
+#
+# This file is distributed under the Clear BSD license.
+# The full text can be found in LICENSE in the root directory.
+
+import os, datetime, pexpect
+import rootfs_boot
+from devices import board, wan, lan, prompt
+from lib.installers import install_snmp
+from cbnlib import now_short
```
mattsm: This is a public repo- we should not put things that depend on a private repo here.

That being said, now_short() can probably move to tests/lib/common.py?

tests/ssid_set_snmp.py
```diff
@@ -0,0 +1,40 @@
...
+            if wifi_device == 'wifi_2g':
+                self.index = '.32'
+                self.set_ssid_name = 'SSIDwifi_2G'
+            elif wifi_device == 'wifi_5g':
+                self.index = '.92'
+                self.set_ssid_name = 'SSIDwifi_5G'
+
+            """Set SSID via SNMP"""
+            ssid_name = self.config.board['station']+self.set_ssid_name
+            wan.sendline("snmpset -v 2c -c private -t 30 -r 3 "+wan_ip+" "+board.mib["wifiMgmtBssSsid"]+self.index+" s "+str(ssid_name))
```
mattsm: Is there a generically defined SNMP mib we can use or is it board specific?

tests/ssid_set_snmp.py
```diff
@@ -0,0 +1,40 @@
...
+    index = None
+    def runTest(self):
+        install_snmp(wan)
+        wan_ip = board.get_interface_ipaddr(board.wan_iface)
+        """"setting ssid for wifi"""
+        wifi_name = ['wifi_2g','wifi_5g']
+        for wifi_device in wifi_name:
+            if wifi_device == 'wifi_2g':
+                self.index = '.32'
+                self.set_ssid_name = 'SSIDwifi_2G'
```
mattsm: These wifi names will need to contain more than a hard coded name. Otherwise we will see so many of these. I suggest using a station prefix with a random SHA as well.

tests/ssid_set_snmp.py
```diff
@@ -0,0 +1,40 @@
...
+            elif wifi_device == 'wifi_5g':
+                self.index = '.92'
+                self.set_ssid_name = 'SSIDwifi_5G'
```
mattsm: same as above

tests/ssid_set_snmp.py
```diff
@@ -24,17 +24,37 @@ def runTest(self):
...
+                set_ssid_name = 'SSIDwifi_2G'
             elif wifi_device == 'wifi_5g':
                 self.index = '.92'
-                self.set_ssid_name = 'SSIDwifi_5G'
-
+                set_ssid_name = 'SSIDwifi_5G'
             """Set SSID via SNMP"""
-            ssid_name = self.config.board['station']+self.set_ssid_name
+            ssid_name = config.board['station']+set_ssid_name
+            wifi_pass = config.wifi_password[0]
```
mattsm: OK.. I think you need to squash these commits down into fewer commits. It's not readable... now I see here you did make the SSID more random.

I don't think we should put the wifi password in the config, we can just use a random one as well.

tests/ssid_set_snmp.py
```diff
@@ -24,17 +24,37 @@ def runTest(self):
...
+            logfile_assert_message(self, idx==0,'Getting the SSID using mib wifiMgmtBssSsid)')
             wan.expect(prompt)
             self.log_to_file += now_short()+"Ssid = %s\r\n" %ssid_name
+            """Get Password via SNMP"""
+            wan.sendline("snmpget -v 2c -c public -t 30 -r 3 "+wan_ip+" "+board.mib["wifiMgmtBssWpaPreSharedKey"]+self.index)
+            idx = wan.expect(['Timeout: No Response from'] + [wifi_pass])
+            logfile_assert_message(self, idx==0,'Getting the SSID using mib wifiMgmtBssWpaPreSharedKey)')
+            wan.expect(prompt)
+        """Apply the settings globally"""
+        wan.sendline("snmpset -v 2c -c private -t 30 -r 3 "+wan_ip+" "+board.mib["wifiMgmtApplySettings"]+'.0'+" i "+'1')
```
mattsm: Again, only generic mibs can go here.

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,91 @@
...
+            wlan.sendline("rm /etc/"+ssid_name+".conf")
+            wlan.expect(prompt)
+            wlan.sendline("wpa_passphrase "+ssid_name+ " >> /etc/"+ssid_name+".conf")
+            wlan.expect("")
+            wlan.sendline(pass_word)
+            wlan.expect(prompt)
+            wlan.sendline("cat /etc/"+ssid_name+".conf")
+            wlan.expect(prompt)
+            wlan.sendline("wpa_supplicant  -B -Dnl80211 -iwlan1 -c/etc/"+ssid_name+".conf")
+            wlan.expect(prompt)
```
mattsm: This needs to be a function on devices/debian_wifi.py and we call it here- not repeated in each test that needs to do it.

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,91 @@
...
+            logfile_assert_message(self, match!=None,'Connection establishment in WIFI')
+            match = re.search(ssid_val,conn_state)
+            logfile_assert_message(self, match!=None,'Connection establishment - SSID verify')
+            '''Ping test'''
+            hostname = "google.com"
+            wlan.sendline("ping -c 1 " + hostname)
+            wlan.expect(prompt)
+            ping_result = wlan.before
+            match = re.search("1 packets transmitted, 1 received, 0% packet loss",ping_result)
+            logfile_assert_message(self, match!=None,'Ping status')
```
mattsm: There is no ping() elsewhere we can reuse?

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,91 @@
...
+        wlan.sendline("ip link show wlan1")
+        wlan.expect(prompt)
+        link_state = wlan.before
+        match = re.search('NO-CARRIER,BROADCAST,MULTICAST,UP',link_state)
+        if match:
+            self.log_to_file += now_short()+"Interface is UP: PASS\r\n"
+        else:
+            wlan.sendline("ip link set wlan1 up")
+            wlan.expect(prompt)
+            self.log_to_file += now_short()+"Interface is set UP\r\n"
```
mattsm: These looks like they should be functions on devices/debian_wifi.py

arguments.py
```diff
@@ -79,7 +79,7 @@ def parse():
     parser.add_argument('--bootargs', metavar='', type=str, default=None, help='bootargs to set or append to default args (board dependant)')
     parser.add_argument('-g', '--golden', metavar='', type=str, default=[], nargs='+', help='Path to JSON results to compare against (golden master)')
     parser.add_argument('-q', '--feature', metavar='', type=str, default=[], nargs='+', help='Features required for this test run')
-
+    parser.add_argument('-pswd', '--password', metavar='', type=str, action="append", default=None, help='password for wifi')
```
mattsm: I don't think this is the right approach. I think we should generate a random password for a sequence of tests and save it on the board object.

tests/check_ssid_wifi.py
```diff
@@ -36,41 +36,72 @@ def WifiTest(self):
...
-            '''Ping test'''
-            self.ping_hostname(ssid_name)
+            '''Check wlan connectivity'''
+            self.wlan_connect(ssid_name)
+
+            '''Ping gateway test'''
+            self.ping_gateway()
+
+            '''Ping hostname test'''
+            self.ping_hostname()
```
mattsm: This looks better... but these functions should be on devices/debian_wifi.py not on the test class.

tests/ssid_set_snmp.py
sravichandran-contractor: It looks like it is board specific because the mib is inside a file named cbnWiFiMgmt.mib

mattsm:
Please rebase/squash. You have lots of commits on top of master at this point:

* 95718a8 (HEAD) tests:check_ssid_wifi: Added the testcase for wifi connectivity
* 0f454e1 devices:debian_wifi: Removed the log_to_file from the function
* 515a6cc Delete snmp_wifi_mibs.py
* 372bcda Delete ssid_set_snmp.py
* 570db56 devices: debian_wifi: Added the wlan container function
* 3863a1b test: ssid_set_snmp: now_short has been removed
* 505f8b3 Update snmp_wifi_mibs.py
* 0ab9a14 New function for the set and get of Wi-Fi mibs
* 9f08b3e Update ssid_set_snmp.py
* 8f91ac8 Update ssid_set_snmp.py
* 7592b41 Update check_ssid_wifi.py
* 3da4b75 Update check_ssid_wifi.py
* 7d8a934 Update wifi.py
* fc3f3dc Update arguments.py
* b1a8b02 Create check_ssid_wifi.py
* 3184864 Update ssid_set_snmp.py
* 92218ac tests: wifi: set unique SSID via SNMP

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) 2017
+#
+# All rights reserved.
+#
+# This file is distributed under the Clear BSD license.
+# The full text can be found in LICENSE in the root directory.
+
+import os, datetime, pexpect, config, re
+import rootfs_boot
+from devices import board, wan, lan, wlan, prompt
```
mgualco-contractor: Nitpicking: wan is unused

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+            self.ssid_name = self.config.board['common']['ssid_name']
+            self.pass_word = self.config.board['common']['pass_word']
+        except:
+            logfile_assert_message (self, False, "ssid_name and password must be specified in the .json 'common' section")
+
+    def runTest(self):
+
+        self.setupTest()
+
+        board.expect(pexpect.TIMEOUT, timeout=50)
```
mgualco-contractor: why this 50s wait?

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+            conn_wlan = wlan.wlan_connect(self.ssid_name)
+            logfile_assert_message(self, conn_wlan!=None,'Connection establishment in WIFI')
+            wlan.sendline("dhclient -v wlan1")
+            wlan.expect(prompt,timeout=40)
+
+            '''Ping gateway test'''
+            wlan.sendline("ifconfig wlan1")
+            wlan.expect(prompt)
+            if_config_res = wlan.before
+            match = re.search(r'inet ([0-9]+\.[0-9]+\.[0-9]+)\.([0-9]+).*',if_config_res)
```
mgualco-contractor: if you are looking for an IP address, could ValidIpv4AddressRegex from regexlib.py be used (just wondering)

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+            logfile_assert_message(self, match!=None,'Ifconfig wlan IP fetch')
+            gateway_ip = match.group(1) + ".1"
+            ip = match.group(2)
+            value = range(2,255)
+            match = re.search(ip,str(value))
+            logfile_assert_message(self, match!=None,'Ifconfig wlan IP verify')
+
+            for host in {gateway_ip,"google.com"}:
+                wlan.sendline("ping -c 1 %s" % host)
```
mgualco-contractor: is a single ping enough?

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+                wlan.expect(prompt)
+                ping_res = wlan.before
+                match = re.search('1 packets transmitted, 1 received, 0% packet loss' , ping_res)
+                logfile_assert_message(self, match!=None,'Ping status')
+
+            '''Kill the wpa supplicant'''
+            wlan.kill_supplicant()
```
mgualco-contractor: killing it twice? here and cleanup/recover?

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+
+            self.cleanupTest()
+
+    def cleanupTest(self):
+        self.recover()
+        logfile_assert_message(self, True, "Nothing to do with cleanup")
```
mgualco-contractor: you already killed the supplicant so you did something. Maybe state "Cleanup finished"

tests/check_ssid_wifi.py
```diff
@@ -0,0 +1,92 @@
...
+
+            '''Generate WPA supplicant file and execute it'''
+            conn_wpa = wlan.wpa_connect(self.ssid_name,self.pass_word)
+            logfile_assert_message(self, conn_wpa!=None,'WPA supplicant initiation')
+
+            '''Check wlan connectivity'''
+            board.expect(pexpect.TIMEOUT, timeout=20)
+            conn_wlan = wlan.wlan_connect(self.ssid_name)
+            logfile_assert_message(self, conn_wlan!=None,'Connection establishment in WIFI')
+            wlan.sendline("dhclient -v wlan1")
```
mgualco-contractor: Just curious, if the link is already up, and an IP already assigned, shouldn't dhclient -r wlan1 be called first?

tests/check_ssid_wifi.py
sravichandran-contractor: I am trying to group only 3 octet from the IP

tests/check_ssid_wifi.py
sravichandran-contractor: Since I use a for loop for 2g and 5g I am using kill supplicant inside the for loop so that multiple instance doesn't run in background.

tests/check_ssid_wifi.py
mattsm: Should probably make a regex that returns an array of all the octet's and you can pick:

".".join(ip[0:3])

sravichandran-contractor:
I have created a new branch to avoid chaos and I will be pushing my commits in the new the branch and initiate a new pull request.

PR #184, opened 2018-12-05
Pr add combined log

mattsm:


mattsm:
6 approvals, yay... still testing some this morning though.

bft
```diff
@@ -296,8 +296,12 @@ def main():
...
         for line in log.split('\r\n'):
             try:
-                ts, text = line.split(' ', 1)
-                combined_list.append({"time": float(ts[1:-1]), "text": text, "name": name})
+                if line is '':
+                    continue
+                if line.startswith(' ['):
+                    line = line[1:]
+                ts, text = line.split(']', 1)
+                combined_list.append({"time": float(ts[1:-1]), "text": text[1:], "name": name})
```
mgualco-contractor: Just nitpicking:
Is the  text[1:] needed? with it the 1st char is removed from the log text line and I see
lan: [607.15287] oot@x1-6-ae-38-3c-b5-05-d6:~# ifconfig eth1
lan: [607.15287] th1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
lan: [607.15287]        inet6 fe80::ac38:3cff:feb5:5d6  prefixlen 64  scopeid 0x20<link>

if i remove the [1:]

lan: [607.15287] root@x1-6-ae-38-3c-b5-05-d6:~# ifconfig eth1
lan: [607.15287] eth1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
lan: [607.15287]         inet6 fe80::ac38:3cff:feb5:5d6  prefixlen 64  scopeid 0x20<link>

mgualco-contractor: i can see you did not have it at first then you added it, probably there is a very good reason...

mattsm: That's not nitpicking, the current way seems wrong... I think I was trying to consume the ']' but the split itself does that and did not check the logs close enough after this change. I will change back.

PR #187, opened 2018-12-06
devices: debian_wifi: Added the below function for wifi

sravichandran-contractor:
wpa_connect - to generate wpa_supplicant with SSID and passphrase and initiate it
 wlan_connect - Check the Wi-Fi connection state in wlan
 link_up - Check the interface is up in wlan
 kill_suuplicant - killing the wpa_supplicant initiated

tests: wifi: Added a return value in the wifi scan function

Signed-off-by: Shakthi Ravichandran <sravichandran.contractor@libertyglobal.com>


devices/debian_wifi.py
```diff
@@ -17,3 +18,54 @@ def scan(self):
...
         return self.before
+
+    def wpa_connect(self,ssid_name,pass_word,wlan_intf):
+
+         '''Generate WPA supplicant file and execute it'''
+
+         self.sendline("rm /etc/"+ssid_name+".conf")
+         self.expect(self.prompt)
+         self.sendline("wpa_passphrase "+ssid_name+ " >> /etc/"+ssid_name+".conf")
+         self.expect("")
```
mgualco-contractor: i am guessing there is no prompt of any kind being displayed....

devices/debian_wifi.py
```diff
@@ -17,3 +18,54 @@ def scan(self):
...
+         self.sendline(pass_word)
+         self.expect(self.prompt)
```
mgualco-contractor: Just a note, i don't really know what the wpa_passphrase does, and it is possible that this  command, NEVER fails. But let's say something has gone wrong. at the moment you are only expecting the prompt, i.e. assuming everything is fine. In this case you could check the value of "$?", if == 0 the command succeeded, !=0 something has failed. 
Would this be applicable in this situation?

devices/debian_wifi.py
```diff
@@ -17,3 +18,54 @@ def scan(self):
...
+         self.sendline("cat /etc/"+ssid_name+".conf")
+         self.expect(self.prompt)
+         self.sendline("wpa_supplicant  -B -Dnl80211 -i"+wlan_intf+ " -c/etc/"+ssid_name+".conf")
+         self.expect(self.prompt)
+         wpa_start = None
+         wpa_start = self.before
+         match = re.search('Successfully initialized wpa_supplicant',wpa_start)
```
mgualco-contractor: Ok, probably if something has gone wrong in the above command, the wpa_supplicant won't initialise.

tests/wifi.py
```diff
@@ -7,3 +7,5 @@ class WifiScan(rootfs_boot.RootFSBootTest):
 
     def runTest(self):
         wlan.scan()
+        scan_output = wlan.scan()
```
mgualco-contractor: do you need to run  wlan.scan() twice?

devices/debian_wifi.py
sravichandran-contractor: Once we pass the "wpa_passphrase" command it waits for password to be entered. Hence in this case we don't expect any prompt.

devices/debian_wifi.py
sravichandran-contractor: In this case we are just passing the password and creating a wpa_passphrase config file. Once the password is provided, it will end up in the prompt. Probably I will add a timeout of 20 sec for expecting the prompt.

devices/debian_wifi.py
sravichandran-contractor: Yes, if something goes wrong wpa_supplicant wont be intialised . As assertion can be thrown in the testcase based on the return value.

tests/wifi.py
sravichandran-contractor: Not sure why this shows twice. The code which I have shows scan.output only once. Will verify and commit it once again.

devices/debian.py
```diff
@@ -869,6 +870,28 @@ def tftp_server_ip_int(self):
         '''Returns the DUT facing side tftp server ip'''
         return self.gw
 
+    def interface_connectivity(self, interface):
+        '''Connection state verify'''
+        self.sendline("iw %s link" % interface)
```
mattsm: This is using the iw command- did I suggest moving this function?

mattsm:
Manually rebased and merged. 

Renamed disconnect_supplicant -> disconnect_wpa for now...

PR #188, opened 2018-12-06
Pr add combined log2

mattsm:


mattsm:
@mgualco-contractor reported this https://gist.github.com/mattsm/c7b4cfb0cf6e32ead3c9866a6f5eeb04

mattsm:
There are hex values in some of those strings we need to account for... looks like the others did not get split properly. Can you share the command / test you ran on slack?

mattsm:
This is still failing... easier repro:

yes | ./bft -b mv1 -x connect -e KernelModules

Still debugging why these lines are not getting split properly....

mgualco-contractor:
the all.log looks clean, apart from the odd  ^@^M 
the prefixing the line with the console# helps a lot. It would be nice if log_to_file sneaked in too :-)

mattsm:
Where is ^@^M

log_to_file will be in the function calls, so they should be there except with extra info for the call itself.

mattsm:
Oh, '^@^M' is in the log? That might be what is really in the log...

We can look at escaping those in the strings here at some point if you can give a more specific example.

mgualco-contractor:
A possible concern,is it possible for all.log to grow too big to be emailed?

mattsm:
Yes, I hope these can be retrieved over Jenkins artifacts or some similar mechanism instead of mailing the full log around.

-M

PR #189, opened 2018-12-12
Modified to install netperf 2.6 and use netserver on wan and netperf …

mkamat:
Revert "Modified to install netperf 2.6 and use netserver on wan and netperf client on lan

### DIFF
--- a/tests/netperf_test.py
+++ b/tests/netperf_test.py
@@ -108,29 +108,60 @@ def run_netperf_tcp(device, run_time, pkt_size, direction="up"):
 class Netperf_UpTCP256(rootfs_boot.RootFSBootTest):
     '''Netperf upload throughput (TCP, packets size 256 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=256)
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
 
 class Netperf_UpTCP512(rootfs_boot.RootFSBootTest):
     '''Netperf upload throughput (TCP, packets size 512 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=512)
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
 
 class Netperf_UpTCP1024(rootfs_boot.RootFSBootTest):
     '''Netperf upload throughput (TCP, packets size 1024 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=1024)
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
 
 class Netperf_DownTCP256(rootfs_boot.RootFSBootTest):
     '''Netperf download throughput (TCP, packets size 256 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=256, direction="down")
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
 
 class Netperf_DownTCP512(rootfs_boot.RootFSBootTest):
     '''Netperf download throughput (TCP, packets size 512 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=512, direction="down")
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
 
 class Netperf_DownTCP1024(rootfs_boot.RootFSBootTest):
     '''Netperf download throughput (TCP, packets size 1024 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=1024, direction="down")
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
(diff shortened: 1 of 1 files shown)


tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
 import time
 import os
 from devices import board, wan, lan, wlan, prompt
-
```
mattsm: Try not to make random whitespace changes.

tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
...
 def install_netperf(device):
     # Check version
     device.sendline('\nnetperf -V')
     try:
-        device.expect('Netperf version 2.4', timeout=10)
+        device.expect('Netperf version 2.6', timeout=10)
```
mattsm: We should probably do 4 or 6... or better yet just 'Netperf version 2.' unless we think it will only work with one version or the other.

tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
...
         device.expect(device.prompt)
     except:
         # Install Netperf
         device.sendline('apt-get update')
```
mattsm: I think you need at least one expect for the prompt in the except: stanza

tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
...
-        device.expect(device.prompt, timeout=60)
-        device.sendline('apt-get -o DPkg::Options::="--force-confnew" -y --force-yes install netperf')
-        device.expect(device.prompt, timeout=60)
-    device.sendline('/etc/init.d/netperf restart')
-    device.expect('Restarting')
-    device.expect(device.prompt)
+        device.expect(device.prompt, timeout=120)
+        device.sendline('apt-get  install netperf')
+        device.expect(device.prompt, timeout=120)
```
mattsm: I think all these lines should be in the except: stanza... maybe the space vs. tabs is showing on the github web page here?

tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
...
 
 class NetperfTest(rootfs_boot.RootFSBootTest):
-    '''Setup Netperf and Ran Throughput.'''
+    '''Setup Netperf and Run Throughput.'''
+    iface_dut = "eth1"
```
mattsm: Ideally we use this on the devices we are using.

lan.iface_dut
wan.iface_dut

This is not an item that is test specific but device specific so we can extract them from those places.

tests/netperf_test.py
```diff
@@ -13,35 +13,40 @@
...
+
     @lib.common.run_once
     def lan_setup(self):
         super(NetperfTest, self).lan_setup()
         install_netperf(lan)
+
```
mattsm: This is OK- but these whitespace changes can go in a separate commit so it's clear one commit is just making whitespace changes. Makes review 100% easier.

tests/netperf_test.py
```diff
@@ -65,24 +70,37 @@ def run_netperf_parse(self, device, timeout=60):
     def run_netperf(self, device, ip, opts="", timeout=60):
         self.run_netperf_cmd(device, ip, opts=opts)
         return self.run_netperf_parse(device)
+   
+    def kill_netserver(self, device):
+        device.sendline("kill -9 `pidof netserver`")
+        device.expect(prompt)
```
mattsm: This looks like the main missing parts for newer versions of netperf... Looks good.

tests/netperf_test.py
```diff
@@ -65,24 +70,37 @@ def run_netperf_parse(self, device, timeout=60):
...
 
     def runTest(self):
         super(NetperfTest, self).runTest()
 
         board.sendline('mpstat -P ALL 30 1')
-        speed = self.run_netperf(lan, "192.168.0.1 -c -C -l 30")
-        board.expect('Average.*idle\r\nAverage:\s+all(\s+[0-9]+.[0-9]+){10}\r\n')
+        board.expect('Linux')
+
+	wan_eth1_addr = wan.get_interface_ipaddr(self.iface_dut)
```
mattsm: wan.iface_dut not self.iface_dut

tests/netperf_test.py
```diff
@@ -65,24 +70,37 @@ def run_netperf_parse(self, device, timeout=60):
...
 def run_netperf_tcp(device, run_time, pkt_size, direction="up"):
+    iface_dut = "eth1"
+    wan_eth1_addr = wan.get_interface_ipaddr(iface_dut)
+    
     if direction == "up":
-        cmd = "netperf -H 192.168.0.1 -c -C -l %s -- -m %s -M %s -D" % (run_time, pkt_size, pkt_size)
+        cmd = "netperf -H %s -c -C -l %s -- -m %s -M %s -D" % (wan_eth1_addr, run_time, pkt_size, pkt_size)
     else:
-        cmd = "netperf -H 192.168.0.1 -c -C -l %s -t TCP_MAERTS -- -m %s -M %s -D" % (run_time, pkt_size, pkt_size)
+        cmd = "netperf -H %s -c -C -l %s -t TCP_MAERTS -- -m %s -M %s -D" % (wan_eth1_addr, run_time, pkt_size, pkt_size)
```
mattsm: I think wan.gw also already has this defined? Not 100% sure.

tests/netperf_test.py
mkamat: used wan.gw instead of  wan.get_interface_ipaddr(self.iface_dut)

tests/netperf_test.py
mkamat: used wan.gw instead of wan.get_interface_ipaddr(self.iface_dut)

tests/netperf_test.py
```diff
@@ -108,29 +108,60 @@ def run_netperf_tcp(device, run_time, pkt_size, direction="up"):
...
 
 class Netperf_DownTCP1024(rootfs_boot.RootFSBootTest):
     '''Netperf download throughput (TCP, packets size 1024 Bytes).'''
     def runTest(self):
+        lib.common.test_msg("Starting netserver on wan...")
+        wan.sendline('/usr/bin/netserver')
+        wan.expect("Starting netserver with host")
         run_netperf_tcp(device=lan, run_time=15, pkt_size=1024, direction="down")
+        wan.sendline("kill -9 `pidof netserver`")
+        wan.expect(prompt)
```
mattsm: Can you refactor start/stop server into functions on the base class and call those instead of repeating the commands each time? It's a wash at this point by lines of code but I think a little bit clearer.

tests/netperf_test.py
```diff
@@ -13,22 +13,22 @@
...
-        device.sendline('apt-get  install netperf')
-        device.expect(device.prompt, timeout=120)
+	device.sendline('apt-get update')
+	device.expect(device.prompt, timeout=120)
+	device.sendline('apt-get  install netperf')
+	device.expect(device.prompt, timeout=120)
 
 class NetperfTest(rootfs_boot.RootFSBootTest):
     '''Setup Netperf and Run Throughput.'''
-    iface_dut = "eth1"
```
mattsm: Ideally these are squashed away.

mattsm:
Merged as https://github.com/lgirdk/boardfarm/commit/886ecdae68a1382e14885123661132d88ccf9879

Closing this PR.

tests/netperf_test.py
mkamat: done

PR #190, opened 2018-12-14
bft: exit with message if unable to find testsuite

mattsm:


mattsm:
@mgualco-contractor please do a quick review.

PR #193, opened 2018-12-19
devices: debian: don't delete tftp files if server is shared

mattsm:
Sometimes we share these devices among multiple other devices esp. when
running as a provisioner so let's not delete all the files

TODO: find out how to cleanup...

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -34,6 +34,7 @@ class DebianBox(base.BaseDevice):
     pkgs_installed = False
     install_pkgs_after_dhcp = False
     is_bridged = False
+    shared_tftp_server = False
 
     iface_dut = "eth1"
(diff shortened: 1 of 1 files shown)


mgualco-contractor:
Just a though, not important, instead of having the shared flag within the "wan-cmts-provisioner", having it  in an option of its own like "wan-shared" so one could use:
"options": "wan-static-ip:<some-ip>, wan-no-eth0, wan-shared", to easily integrate it with current systems. Would see that as doable ?

mattsm:
I see your point but wan-cmts-provioner is sort of an aberration. It will go away soon and we will have a provisioner and a wan node for each CM. Right now this node is shared among all SVL CMs so it's bad.

PR #194, opened 2018-12-19
Pr gui ss ipv6 updates

mattsm:


devices/debian.py
```diff
@@ -796,6 +797,10 @@ def start_lan_client(self, wan_gw=None):
         self.expect(self.prompt)
         self.sendline("sed 's/#Allow/Allow/g' -i /etc/tinyproxy.conf /etc/tinyproxy/tinyproxy.conf")
         self.expect(self.prompt)
+        self.sendline("sed '/Listen/d' -i /etc/tinyproxy.conf /etc/tinyproxy/tinyproxy.conf")
+        self.expect(self.prompt)
+        self.sendline('echo "Listen ::" >> /etc/tinyproxy.conf')
```
mgualco-contractor: so you are first uncommenting the Allow's, then you delete the line with Listen (i think is the "#Listen 192.168.0.1") in the tinyproxy directory. Then you append the "Listen ::" to the tinyproxy.conf  in the /etc directory.
Is this intentional? Is it somehow related to the fixing you are doing to the ipv6 at your location?

mattsm: I'm deleting all Listen lines and adding the one back that I want....

This probably does not fix ipv6 as we don't have ipv6 from the docker host into the docker container, and the way proxies work (I think so far) is that to use IPV6 it needs to have IPV6 all the way to the machine running the proxy. 

I need to investigate this... if it can't be fixed with the proxy or docker we need to run an alternate proxy like ssh dynamic proxy

mgualco-contractor: The reason why I am asking is that by looking at the commands: 

sed -i 's/^Port 8888/Port 8080/' /etc/tinyproxy.conf /etc/tinyproxy/tinyproxy.conf
sed 's/#Allow/Allow/g' -i /etc/tinyproxy.conf /etc/tinyproxy/tinyproxy.conf
sed '/Listen/d' -i /etc/tinyproxy.conf /etc/tinyproxy/tinyproxy.conf
echo "Listen ::" >> /etc/tinyproxy.conf

the last command appends the "Listen ::"  to an empty file, i.e.  /etc/tinyproxy.conf, I thought the "Listen ::" should probably go in the /etc/tinyproxy/tinyproxy.conf
unless i am reading this wrong (which is very possible as I might be in holiday mode still)

mgualco-contractor: I SEE, now it makes sense. The machine I checked did not have the symlink, but 2 different files. Ok ok , now it makes sense.... Thx

tests/screenshot_gui.py
```diff
@@ -50,6 +50,26 @@ def runTest(self):
...
         print("Using proxy %s" % proxy)
         driver = lib.common.get_webproxy_driver(proxy)
+
+        return driver
+
+    def runTest(self):
+        driver = self.start_browser()
+
+        print("Browser is running, connect and debug")
+        print("Press Control-[ to exit interactive mode")
```
mgualco-contractor: isn't Control-] to exit interact mode?

mattsm: Will check that too before merging this...

devices/debian.py
mattsm: I was somewhat wrong- address in a followup commit... now I just run all the commands on both files... if they don't exist it won't catch an error... needs some more love later..

tests/screenshot_gui.py
mattsm: Fixed in commit

PR #196, opened 2019-01-02
tests: screenshot_gui.py: pr_add_enable_cpe_service_for_ScreenshotGUI

luketseng:
- add enable mib of cmConfigCpeServiceEnable to get GUI via lan port.


tests/screenshot_gui.py
```diff
@@ -48,6 +48,12 @@ def runTest(self):
             print e
             raise Exception("No reasonable http proxy found, please add one to the board config")
 
+        wan_ip=board.get_interface_ipaddr(board.wan_iface)
+        wan.sendline("snmpset -v 2c -c public -t 30 %s %s.0 x 0x40" %(wan_ip, board.mib["cmConfigCpeServiceEnable"]))
+        assert 0 != wan.expect(['Timeout: No Response from'] + ['STRING: \"@\".*\r\n'])
+        wan.expect(prompt)
+        board.expect(pexpect.TIMEOUT, timeout=5)
+
```
mattsm: This is board specific... should not go in open source repo.

If you want to add a function on board class

board.enable_mgmt_gui()

And have a blank function as the default on the parent class we can add that function call here.

luketseng: I modified it.
Please review commit on gerrit(with 652), thx.

mattsm: This part of the change looks good but you need to add a default empty function so all other boards running this code will not fail.

in devices/openwrt_router.py (this needs to be refactored but this is misnamed slightly for now)

def enable_mgmt_gui(self):
    pass

def

luketseng: I so confuse.
I add code in mv1.py and call this function in ScreenshotGUI.
How can I fit your comment? Please talk more clear, thx.
```
    def enable_mgmt_gui(self, board, wan):
        '''enable mv1 gui via lan port'''

        wan_ip=board.get_interface_ipaddr(board.wan_iface)
        wan.sendline("snmpset -v 2c -c public -t 30 %s %s.0 x 0x40" %(wan_ip, board.mib["cmConfigCpeServiceEnable"]))
        assert 0 != wan.expect(['Timeout: No Response from'] + ['STRING: \"@\".*\r\n'])
        wan.expect(wan.prompt)
        board.expect(pexpect.TIMEOUT, timeout=5)
```

luketseng: like as below in ScreenshotGUI class?

self.enable_mgmt_gui()

def enable_mgmt_gui(self):
   try:
       board.enable_mgmt_gui(board, wan)
   except:
       pass

mattsm: No, you want a stub function in the openwrt class that does nothing. Not a try/except in the test class.

luketseng: So I only add as below to openwrt class, is right?
it can let board.enable_mgmt_gui(board, wan) do nothing if running RDK-B?

```
def enable_mgmt_gui(self):
    pass
```

luketseng: Done, and please review mv1.py modify from gerrit

mattsm: The last change looks good. I will merge soon.

PR #197, opened 2019-01-03
tests: cdrouter_bootstrap: modified cdrouter provision setup

MartinCBN:
modified cdrouter provison setup
add recover dut to CM_online_freq

Signed-off-by: MartinCBN <martin_wu@compalbn.com>

### DIFF
--- a/tests/cdrouter_bootstrap.py
+++ b/tests/cdrouter_bootstrap.py
@@ -13,7 +13,7 @@ from cdrouter.packages import Package
 
 import time
 import rootfs_boot
-from devices import board, wan, lan, wlan, prompt
+from devices import board, wan, lan, wlan, prompt, provisioner
 import os
 import pexpect
 import lib
@@ -57,6 +57,16 @@ class CDrouterStub(rootfs_boot.RootFSBootTest):
             wan.expect(prompt)
 
         c = CDRouter(self.cdrouter_server)
+
+        """Go to cdrouter frequency and reboot to Dual-stack config"""
+        wan_ip = board.get_interface_ipaddr("wan0")
+        if self.config.board['cdrouter_location'] == 'CBN':
+            cm_freq = self.config.board['cdrouter_freq']
+        else:
+            cm_freq = self.config.board['common']['CM_online_freq']
+        wan.sendline("snmpset -v 2c -c public -t 30 -r 3 "+wan_ip+" "+board.mib["lastKnownFreq"]+".0 i "+cm_freq)
+        assert 0 != wan.expect(['Timeout: No Response from'] + prompt)
+        board.reprovision(provisioner, cm_cfg="9_EU_CBN_Dual-Stack_LG.txt", mta_cfg="AutoCA.txt")
 
         try:
             board.sendcontrol('c')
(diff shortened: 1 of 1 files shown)


tests/cdrouter_bootstrap.py
```diff
@@ -323,6 +334,15 @@ def recover(self):
...
             board.wait_for_linux()
 
+        wan_ip = board.get_interface_ipaddr(board.wan_iface)
+        cm_freq = self.config.board['common']['CM_online_freq']
+        wan.sendline("snmpset -v 2c -c public -t 30 -r 3 "+wan_ip+" "+board.mib["lastKnownFreq"]+".0 i "+cm_freq)
+        assert 0 != wan.expect(['Timeout: No Response from'] + prompt)
+        board.enter_arm_shell()
+        board.arm.sendline('reboot')
+        board.wait_for_linux()
+        board.wait_for_network()
```
mattsm: This is not generic, board specific stuff can't go here....

tests/cdrouter_bootstrap.py
```diff
@@ -58,6 +58,16 @@ def runTest(self):
 
         c = CDRouter(self.cdrouter_server)
 
+        """Go to cdrouter frequency and reboot to Dual-stack config"""
+        wan_ip = board.get_interface_ipaddr("wan0")
+        if self.config.board['cdrouter_location'] == 'CBN':
```
mattsm: I don't think using the location is the right thing to flag off of here.

tests/cdrouter_bootstrap.py
```diff
@@ -58,6 +58,16 @@ def runTest(self):
...
+            cm_freq = self.config.board['cdrouter_freq']
+        else:
+            cm_freq = self.config.board['common']['CM_online_freq']
```
mattsm: I thought we talked with @SyHuangCBN about this? Can't we drop this freq from the config and determine it dynamcially? (From a mac domain listed in the board config?)

tests/cdrouter_bootstrap.py
```diff
@@ -58,6 +58,16 @@ def runTest(self):
...
+        wan.sendline("snmpset -v 2c -c public -t 30 -r 3 "+wan_ip+" "+board.mib["lastKnownFreq"]+".0 i "+cm_freq)
+        assert 0 != wan.expect(['Timeout: No Response from'] + prompt)
+        board.reprovision(provisioner, cm_cfg="9_EU_CBN_Dual-Stack_LG.txt", mta_cfg="AutoCA.txt")
```
mattsm: This is all board specific... it need to be generic. If you need to do something board specific pre and post run you can add a hook... or possibly just having a test that inherits and runs the provisioning code before starting CDrouter...
```
class myCDrouterTest(CDrouterBootStrap):
   def runTest(self):
       board.provision()
       super(runTest())
```

tests/cdrouter_bootstrap.py
```diff
@@ -176,21 +186,22 @@ def add_cdrouter_config(config):
...
+    testvar wanIspAssignIp %s
+    testvar wanNatIp %s
+    testvar IPv4HopCount %s
+    testvar lanDnsServer %s
+    testvar wanDnsServer %s""" % (self.config.board['cdrouter_wanispip'], \
+                                  self.config.board['cdrouter_wanispgateway'], \
+                                  wan_ip, wan_ip, \
+                                  self.config.board['cdrouter_ipv4hopcount'], \
+                                  board.get_dns_server(), \
+                                  board.get_dns_server_upstream())
```
mattsm: These need to be defined in the boardfarm config for CBN... not just overridden. They should also be removed from configs/cdrouter-mv1.conf only generic items should live in that config.

mattsm:
Any followup- can't add board specific items here.

tests/cdrouter_bootstrap.py
MartinCBN: Understood, running the provisioning code before testing cdrouter.

tests/cdrouter_bootstrap.py
MartinCBN: This location is used for testing cdrouter with another way. The NTL team(maual test team) using cdrouter to go to specific frequency to test case, not we usually using frequency. Therefore, I just using this location to change frequency. Is it proper?

tests/cdrouter_bootstrap.py
MartinCBN: Okay, I will modified this part after mac domain listed in the board config

mattsm:
This is old, going to close... need to move cdrouter to device model to take changes like this...

PR #204, opened 2019-01-14
tests: lib: common: add helper function to dynamically create a new test

mattsm:
This function will insert a new test class into the globals() so bft
will see it and be able to execute it. It's typical use is to have a
base class as a parent that does the brunt of the work and some
attributes we can add to make it behave differently.

You can though add functions as attributes and build tests dynamically
that way as well.


mattsm:
This does not work right when in a separate function, so we need to call it locally- more tests later will be added that use this.

PR #212, opened 2019-01-23
lib: netork_helper.py: For mib get IP and MAC value format

GerbenChenCBN:
For Snmpwalk mib value get likes:
Mib and ACS get value are different
example:
00 00 00 00 00 00 -> 00:00:00:00:00:00
or
00 00 00 00 -> 00:00:00:00

Signed-off-by: gerben <gerben_chen@compalbn.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,4 @@ pyvirtualdisplay
 Faker
 zeep
 matplotlib
+netaddr
--- a/tests/lib/network_helper.py
+++ b/tests/lib/network_helper.py
@@ -1,0 +1,23 @@
+from netaddr import *
+import re
+import ipaddress
+def mac_to_snmp_format(mac_addr):
+    mac_tmp = re.sub("[\s\.\-]","",mac_addr)
+    mac = EUI(mac_tmp,dialect=mac_unix)
+    mac_final = str(mac).upper()
+    return mac_final
+
+
+def ipv4_to_snmp_format(ipv4_str):
+    ipv4_tmp = re.sub("[\s\.\-]","",ipv4_str)
+    ipv4_decimal = int(ipv4_tmp,16)
+    ipv4_format = ipaddress.IPv4Address(ipv4_decimal)
+    ipv4_address = ipaddress.ip_address(ipv4_format)
+    return ipv4_address
+
+def ipv6_to_snmp_format(ipv6_str):
+    ipv6_tmp = re.sub("[\s\.\-]","",ipv6_str)
+    pattern = re.compile('.{4}')
+    ipv6_tmp_ip = ':'.join(pattern.findall(ipv6_tmp))
+    ipv6_address = ipaddress.ip_address(ipv6_tmp_ip)
+    return ipv6_address


mattsm:
@ketantewari @mgualco-contractor what do you think of this?

ketantewari:
Format for MAC ( xx xx xx xx xx xx ) in MIBs make sense as you can see in below output:
$ snmpwalk -c private -v2c 10.15.16.19 .1.3.6.1.2.1.17.4.3.1.1
SNMPv2-SMI::mib-2.17.4.3.1.1.136.247.199.41.53.78 = Hex-STRING: 88 F7 C7 29 35 4E

The output format of MAC address in ACS differs:
As per standard (tr-181-2-6-0):
MAC Address String (17) : All MAC addresses are represented as strings of 12 hexadecimal digits (digits 0-9, letters A-F or a-f) displayed as six pairs of digits separated by colons.

Works in Axeros 
Device.DeviceInfo.X_LGI-COM_CM_MAC : AC:22:05:9E:19:6F

But there are ACSs with different output format (depending on config):
XX-XX-XX-XX-XX-XX (IETF format ) Cisco ACS
XX XX XX XX XX XX (unformatted) - Friendly ACS

CMTS MAC format : XXXX.XXXX.XXXX 
Don't know if this format is already considered.

mattsm:
We need to be able to handle all these cases. I suggest something similar the IP regex to match, and once they are parsed out via the regex we can encode them to any format from that point.

mgualco-contractor:
we already have CmtsMacFormat='([0-9a-f]{4}\.[0-9a-fA-F]{4}\.[0-9a-f]{4})'  in the regex lib.
the others should be easy (cough) to add.
Probably some one liners conversion utils would be good...

mattsm:
Maybe the appropriate method is to have the CMTS init function set the appropriate regex's as attributes?

```
from lib.regex import cmts_mac_6hexandadot

def init(...)
     self.mac_regex = cmts_mac_6hexandadot

    ...

```

GerbenChenCBN:
@mattsm 
I use regex to set snmp mib mac e.g xx xx xx xx xx xx.

SnmpMibMacFormat = '([0-9a-f]{2}\s[0-9a-f]{2}\s[0-9a-fA-F]{2}\s[0-9a-fA-F]{2}\s[0-9a-f]{2}\s[0-9a-fA-F]{2}'

But I think I also need to use replace let space become ':' and return, beacuse I need the mac to in comparison with acs value.

Maybe have other ways to slove this problem, did you have any propose for reference?

mattsm:
Can we use netaddr library for storing mac addresess?

https://netaddr.readthedocs.io/en/latest/tutorial_02.html

-M

mattsm:
There is even a mac cisco dialet:

>>> mac.dialect = mac_cisco

mattsm:
And you can create your own dialects...

GerbenChenCBN:
@mattsm @ketantewari  @mgualco-contractor  
How did you think about this function?

tests/lib/network_helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+'''IP Decimal Type 2, 10, 16
+e.g.
+ip2addr = '10101100000100000000000100000001'
+ip10addr = '2886729985'
+ip16addr = 'AC100101'
+ip = ipv4_format(ip$addr,$decimal)
+also turn to 172.16.1.1
+'''
+def ipv4_format(ipv4_str,ip_decimal):
```
mattsm: I can't quite tell the purpose of this function can you please say what you are converting it too?

ip_to_dotted_decimal()

Secondly, why not use the IP address library the rest of the boardfarm is trying to standardize on?

https://docs.python.org/3/howto/ipaddress.html

tests/lib/network_helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+from netaddr import *
+import socket
+import struct
+import re
+class mac_custom(mac_unix):
+    pass
+
+def mac_format(mac_addr):
+    mac_tmp = re.sub("[\s+\.\!\-\:\/_,$%^*(+\"\')]+|[+()?【】“”！，。？、~@#￥%……&*（）]+',","",mac_addr)
```
mattsm: We need to be explicit about what the input and output format is, if not we can add a try / except or for loop to try to find the right one... but this is too generic.

Ideally we just use word_sep in the class above for the cases we want to detect.

tests/lib/network_helper.py
```diff
@@ -0,0 +1,36 @@
...
+    mac_custom.word_fmt = '%.2X'
```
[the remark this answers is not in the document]
GerbenChenCBN: I'm sorry, I try code forget to delete it.
And I will add upper() inside it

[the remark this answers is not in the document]
mattsm: No no... if you NEED it it's fine! I was just making sure you were doing this because you needed it. This example was verbatim from the documentation so I was not sure.

tests/lib/network_helper.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+from netaddr import *
+import socket
+import struct
+import re
+class mac_custom(mac_unix):
+    pass
+
+def mac_format(mac_addr):
```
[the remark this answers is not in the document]
mattsm: Now if you need the upper case, it should be called mac_to_snmp_fmt()

tests/lib/network_helper.py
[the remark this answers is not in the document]
GerbenChenCBN: I look lib to check, it was repeat, so I modify next patch and test.
It work same as before.
Then, I try it use upper and lower, it also could do, but I think almost mac type are upper, it can be eazy to use at future.

tests/lib/network_helper.py
GerbenChenCBN: Lib only to check, I also need to deal with ip-address again. likes:
ipaddress.ip_network(3221225984)
get 192.0.2.0/32
I need to delete '/32'.

GerbenChenCBN:
@mattsm I modify it for your suggest, help me to check it, thanks.

tests/lib/network_helper.py
```diff
@@ -0,0 +1,34 @@
...
+e.g.
+ip2addr = '10101100000100000000000100000001'
+ip10addr = '2886729985'
+ip16addr = 'AC100101'
+ip = ipv4_format(ip$addr,$decimal)
+also turn to 172.16.1.1
+'''
+def ipv4_to_snmp_format(ipv4_str,ip_decimal):
+    type_list = [2,10,16]
+    ipv4_format = re.sub("[\s+\.\!\-\:\/_,$%^*(+\"\')]+|[+()?【】“”！，。？、~@#￥%……&*（）]+',","",ipv4_str)
```
mattsm: Can you please leave a comment why this regex is required? Why do  you have a 

+"," at the end of the first argument instead of just adding it to the original string?

tests/lib/network_helper.py
```diff
@@ -0,0 +1,34 @@
...
+    if ip_decimal in type_list:
+        ipv4_address = socket.inet_ntoa(struct.pack('I',socket.htonl(int(ipv4_format,ip_decimal))))
```
mattsm: Maybe use ipaddress.IPv4Adress for this going foward? Not requiring this change but I want to use this library to represent all IP addresses going forward

GerbenChenCBN: I use ipaddress model to try to test, but I have a problem.
How can I use Hex to Decimal ? I can not Found function to solve it.
Maybe I could add ipaddress.IPV4LENGTH to do something under line 31 ?

mattsm: I think we can just leave this as is for now- do you have an update for the PR?

mattsm:
Can you take a look at this and let me know what you think?

https://review.gerrithub.io/c/lgirdk/boardfarm/+/448492

GerbenChenCBN:
@mattsm 
https://review.gerrithub.io/c/lgirdk/boardfarm/+/448492
This is first version. 
Why it different with here? https://review.gerrithub.io/c/lgirdk/boardfarm/+/12

I try other method about ipaddress, I will upload later.

tests/lib/network_helper.py
GerbenChenCBN: @mattsm 
I upload new patchset, it delete socket and using ipaddress lib to check.

mattsm:
> @mattsm
> https://review.gerrithub.io/c/lgirdk/boardfarm/+/448492
> This is first version.
> Why it different with here? https://review.gerrithub.io/c/lgirdk/boardfarm/+/12
> 
> I try other method about ipaddress, I will upload later.

This your changed rebased and cleaned up...

requirements.txt
```diff
@@ -18,3 +18,4 @@ pyvirtualdisplay
 Faker
 zeep
 matplotlib
+netaddr
```
mattsm: GitHub does not show it, but there is no newline at the end of this file...

tests/lib/network_helper.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+from netaddr import *
+import re
+import ipaddress
+def mac_to_snmp_format(mac_addr):
+    if mac_addr != None:
+        try:
+            mac_tmp = re.sub("[\s+\.\!\-\:\/_,$%^*(+\"\')]+|[+()?【】“”！，。？、~@#￥%……&*（）]+',","",mac_addr)
```
mattsm: I've asked before, why are you doing this? I don't like you stripping out all this because it might causing us to pass and convert something we did not intend to convert. Please do the conversion you need and just that...

tests/lib/network_helper.py
```diff
@@ -0,0 +1,41 @@
...
+            mac = EUI(mac_tmp,dialect=mac_unix)
+            mac_final = str(mac).upper()
+            return mac_final
+        except:
+            assert False , 'Mac Type is Error!'
+    else:
+        assert False , 'Error, MAC is None!'
```
mattsm: Did you look at my reworked version?

https://review.gerrithub.io/c/lgirdk/boardfarm/+/448492/3/tests/lib/network_helper.py

you don't need to be cathcing all these, if the parse fails this function should raise an exception we should not be catching it and just raising a new exception with our own messages...

tests/lib/network_helper.py
```diff
@@ -0,0 +1,41 @@
...
+    if len(ipv6_tmp) == 32:
+        pattern = re.compile('.{4}')
+        ipv6_tmp_ip = ':'.join(pattern.findall(ipv6_tmp))
+        try:
+            ipv6_address = ipaddress.ip_address(u'%s' %ipv6_tmp_ip)
+            return ipv6_address
+        except:
+            assert False , 'IPv6 Address Type is Error!'
+    else:
+        assert False , 'IPv6 Address Type is Error!'
```
mattsm: Same comments as above, this is quite complex for what it's doing... these two line functions do the same thing:

https://review.gerrithub.io/c/lgirdk/boardfarm/+/448492/3/tests/lib/network_helper.py#14

tests/lib/network_helper.py
GerbenChenCBN: Okay, so I only use what I want.

tests/lib/network_helper.py
GerbenChenCBN: Okay, I just use function but not assert, right?

mattsm: Yes, please review the link I sent...

GerbenChenCBN:
@mattsm  I have upload new patch, help me review please.

mattsm:
Merged manually after a few formatting fixes. Please review them and take note for future commits.

PR #216, opened 2019-01-26
Pr match tun0 interface

mgualco-contractor:
in get_interface_ipaddr() added match for
inet 10.8.0.1  netmask 255.255.255.0  destination 10.8.0.1

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -15,6 +15,7 @@ import os
 import binascii
 import glob
 import ipaddress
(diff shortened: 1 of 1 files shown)


mattsm:
Can you please drop the regex for the uptime from this PR? I think it's broken...

mgualco-contractor:
> Can you please drop the regex for the uptime from this PR? I think it's broken...

will do

mgualco-contractor:
squashed

PR #217, opened 2019-01-28
devices: __init__.py : fix bug for cbn.json lan2 can not import

GerbenChenCBN:
add lan2 in __init__.py to fix lan2 import fail from cbn.json
it use on 2 CPEs from "http://172.16.1.237:8080/c/boardfarm-cbn/+/1716"

Signed-off-by: gerben <gerben_chen@compalbn.com>

### DIFF
--- a/devices/__init__.py
+++ b/devices/__init__.py
@@ -20,7 +20,7 @@ wlan = None
 wlan2g = None
 wlan5g = None
 prompt = None
-
+lan2 = None
 device_files = glob.glob(os.path.dirname(__file__)+"/*.py")
 if 'BFT_OVERLAY' in os.environ:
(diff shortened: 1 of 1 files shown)


GerbenChenCBN:
@mattsm this fix need to review first. It about our daily test

mattsm:
I think we need generic way to handle this better... I wonder if we can inspect a class (or a test file) and see what devices it will import and skip that test instead... For now I think one should just delay the import of lan2 until run time (e.g. move it into the runTest() function)

mattsm:
@mgualco-contractor can you move the lan2 import to runTest to fix this for now?

mgualco-contractor:
> @mgualco-contractor can you move the lan2 import to runTest to fix this for now?

done

PR #223, opened 2019-01-30
tests: rootfs_boot: only route traffic if we are a debian based prov

mattsm:


lynnlincbn:
Test passed, please help to merge it, thanks.

mattsm:
At some point this needs to be refactored into a function on the prov class, and stuffed out for prov that are not involved in routing.

PR #224, opened 2019-01-31
devices: __init__: added check for kermit installation

mgualco-contractor:
There may be a chance in the future for kermit to be used
instead of telnet (as the latter seem to be taking 100% cpu for
no apparent reason!)

### DIFF
--- a/devices/__init__.py
+++ b/devices/__init__.py
@@ -9,6 +9,7 @@ import sys
 import glob
 import inspect
 import pexpect
+import termcolor
 
 # insert tests lib so devices and tests can share the same libraries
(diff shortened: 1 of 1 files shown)


devices/__init__.py
```diff
@@ -42,6 +43,18 @@
             device_mappings[device_file].append(ref)
             exec("from %s import %s" % (x, obj))
 
+def check_for_cmd_on_host(cmd):
+    from distutils.spawn import find_executable
+    if find_executable(cmd) is None:
```
mgualco-contractor: Try
from common import cmd_exists

PR #230, opened 2019-02-05
tests: lib: installers: added helper function "install_dovecot" to in…

premandfriends:
…stall dovecot server

	- Helper function will install Dovecot core server which runs both POP3 and IMAP

Signed-off-by: Prem Kumar K N <prekumar.contractor@libertyglobal.com>

### DIFF
--- a/tests/lib/installers.py
+++ b/tests/lib/installers.py
@@ -358,6 +358,22 @@ def install_IRCserver(device):
         device.expect(device.prompt)
     except:
         device.expect(device.prompt)
+        device.sendline('apt-get install update-inetd -y') # Update inetd before installation
+        device.expect(device.prompt, timeout=90)
         device.sendline('apt-get install inspircd -y')
         device.expect(['Setting up inspircd'], timeout=90)
+        device.expect(device.prompt)
+
+def install_dovecot(device):
+    '''Install dovecot server if not present.'''
+    device.sendline('dovecot --version')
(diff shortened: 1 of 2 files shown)


tests/lib/installers.py
```diff
@@ -358,6 +358,22 @@ def install_IRCserver(device):
         device.expect(device.prompt)
     except:
         device.expect(device.prompt)
+        device.sendline('apt-get install update-inetd -y') # Update inetd before installation
+        device.expect(device.prompt, timeout=90)
```
mattsm: You meant to include this as well? These make sense as two separate commits or at least add the fact you are fixing an issue with IRC installation as well in the commit message.

premandfriends: Update the commit message Matt . Hope this makes sense now.

mattsm: No the commit does not make sense, it says nothing about why we are updating the IRC function.

Also if you could make the lines 80 chars or less....

mattsm:
I'm not sure you meant to add that new commit?

tests/lib/installers.py
premandfriends: Sorry for the Mess Matt. I have address your comments and combined the 2 changes and committed.
https://github.com/lgirdk/boardfarm/commit/185b20dfd7d20827af346905fe5116ee24c4cd0a

PR #236, opened 2019-02-11
tests: lib: installers: the openvpn-install.sh is no longer curled

mgualco-contractor:
Now use the committed openvpn-install.sh instead of curling it
every time. The curl line was left for reference (commented out).


mgualco-contractor:
do not merge just yet

mgualco-contractor:
squashed and ready to be merged

PR #239, opened 2019-02-12
devices: debian: set_link_status() - To set the link status

sravichandran-contractor:
devices: debian_wifi: - Function to connect wifi
    wlan_ssid_connect() - connect to wifi in security disabled mode
    wlan_ssid_disconnect() - disconnect wifi in security disabled mode
    start_lan_client() - Function to call the DHCP process
    linux_wifi_client() - generic function which scan ssid and verifies wifi connectivity

lib: common: Functions for file open, snmp set and get
    file_open() - function to open a file and read the values
    snmp_mib_set() - SNMP set function to set values
    snmp_mib_get() - SNMP get function to get values

lib: wifi: Generic snmp set and get functions for wifi
    wifi_stub() - Generic class for wifi functions

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -966,6 +966,11 @@ EOF''' % (self.iface_dut, self.iface_dut, self.iface_dut))
         else:
             return None
 
+    def set_link_state(self, interface, state):
+        '''Setting the interface status'''
+        self.sendline("ip link set %s %s" % (interface,state))
+        self.expect(self.prompt)
+
 if __name__ == '__main__':
     # Example use
     try:
--- a/devices/debian_wifi.py
+++ b/devices/debian_wifi.py
@@ -1,7 +1,8 @@
-
 import re
 import debian
-
+import pexpect
+from devices import wlan, board, prompt
+from lib.regexlib import ValidIpv4AddressRegex
 
 class DebianWifi(debian.DebianBox):
     '''Extension of Debian class with wifi functions'''
@@ -53,3 +54,38 @@ class DebianWifi(debian.DebianBox):
     def disconnect_wpa(self):
         self.sendline("killall wpa_supplicant")
         self.expect(self.prompt)
+
+    def wlan_ssid_connect(self, ssid_name):
+        self.sendline("iwconfig %s essid %s" % (self.iface_wlan,ssid_name))
+        self.expect(self.prompt)
+
+    def wlan_ssid_disconnect(self):
+        self.sendline("iw dev %s disconnect" % self.iface_wlan)
+        self.expect(self.prompt)
+
+    def start_lan_client(self, wan_gw=None):
+        self.iface_dut = self.iface_wlan
+        super(DebianWifi, self).start_lan_client()
+
+    def linux_wifi_client(self, board, ssid_name, password=None):
+        '''Scan for SSID and verify connectivity'''
+        output = self.scan()
+        match = re.search("%s" % ssid_name,output)
+        assert match!=None,'SSID value check in WLAN container'
+
+        link = self.link_up(self.iface_wlan)
+        if link==None:
+            wlan.set_link_state(self.iface_wlan)
+
+        if password:
+            conn_wpa = self.wpa_connect(ssid_name,password)
+            assert conn_wpa!=None,'WPA supplicant initiation'
+        else:
+            self.wlan_ssid_connect(ssid_name)
+
+        board.expect(pexpect.TIMEOUT, timeout=20)
+        conn_wlan = self.wlan_connectivity()
+        assert conn_wlan!=None,'Connection establishment in WIFI'
+
+        self.start_lan_client()
+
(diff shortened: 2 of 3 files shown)


devices/debian_wifi.py
```diff
@@ -53,3 +54,38 @@ def wlan_connectivity(self):
...
+    def wlan_ssid_connect(self, ssid_name):
+        self.sendline("iwconfig %s essid %s" % (self.iface_wlan,ssid_name))
+        self.expect(self.prompt)
+
+    def wlan_ssid_disconnect(self):
+        self.sendline("iw dev %s disconnect" % self.iface_wlan)
+        self.expect(self.prompt)
+
+    def start_lan_client(self, wan_gw=None):
+        self.iface_dut = self.iface_wlan
```
mattsm: Do we need iface_wlan at all? Should just use iface_dut instead?

devices/debian.py
```diff
@@ -966,6 +966,11 @@ def link_up(self, interface):
         else:
             return None
 
+    def set_link_status(self, interface, status):
```
mattsm: Can we rename this to set_link_state? This is just turning the interface on and off, right?

tests/lib/wifi.py
```diff
@@ -177,3 +178,165 @@ def uciSetWifiSecurity(board, vap_iface, security):
...
+    def set_password_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def enable_channel_util_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def set_operating_mode_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def set_bandwidth_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def set_channel_no_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
```
mattsm: This class looks ok- but why is everything appended with _verify?

tests/lib/wifi.py
```diff
@@ -177,3 +178,165 @@ def uciSetWifiSecurity(board, vap_iface, security):
...
+       return mib_out
+    else:
+       return None
+
+def get_broadcast(device, board, iface_ip, mib_name, index):
+    mib_out = snmp_mib_get(device,board,iface_ip,mib_name,index)
+    if mib_out != None:
+        return mib_out
+    else:
+        return None
```
mattsm: I don't think we need these functions. I think the device class handler for wifi_snmp() can just call these. These might not be generic enough for other board types anyways so we will just let the device classes handle it all.

tests/lib/common.py
```diff
@@ -389,3 +392,64 @@ def configure_postfix(device):
...
+            if "json" in extension:
+                read_data = json.load(f)
+            elif "txt" in extension:
+                read_data = f.read()
+            return read_data
+        except Exception as e:
+            print "Could not read file from the path:", file_path
+            raise (e)
+
+def snmp_mib_set(device, board, iface_ip, mib_name, index, set_type, set_value, timeout=10, retry=3):
```
mattsm: These SNMP functions are OK- but they really could be in their own commit so we could let them through and not be held up by the other reviews.

tests/lib/common.py
```diff
@@ -389,3 +392,64 @@ def configure_postfix(device):
...
+    with open(file_path) as f:
+        try:
+            if "json" in extension:
+                read_data = json.load(f)
+            elif "txt" in extension:
+                read_data = f.read()
+            return read_data
+        except Exception as e:
+            print "Could not read file from the path:", file_path
+            raise (e)
```
mattsm: Can we keep this a bit simplier for now:
```
def file_open_json(file):
    with open(file) as f:
        return json.load(f)
```

We should throw an exception if something fails, since we're only loading a json file.

Also, this could and should be in it's own commit.

devices/debian.py
sravichandran-contractor: Sure will rename the same

tests/lib/wifi.py
sravichandran-contractor: Since the function in device class does a set and verify of the values we appended with _verify. The same name has been used here. Should we remove this?

tests/lib/common.py
sravichandran-contractor: Agreed will change the same

devices/debian_wifi.py
sravichandran-contractor: If I use iface_dut it is taking eth0 as default. As I wasn't the interface to be wlan1 am passing the iface_wlan1 here

tests/lib/wifi.py
sravichandran-contractor: Sure. Will dissolve this function and will call snmp set and get directly in the device class handler for wifi_snmp()

mattsm:
Merged manually

PR #242, opened 2019-02-14
tests: lib: gui_helper module added

kvetrivel-contractor:
Added gui functions to set and get the values for text box , radio button
and drop down

Signed-off-by: Kalpana Vetrivel <kvetrivel.contractor@libertyglobal.com>


mattsm:
Merged manually.

PR #250, opened 2019-02-21
device: debian: retrieves a file form the tftp to the localhost

mgualco-contractor:
retrieves a file form the tftp server to the localhost.

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -36,6 +36,7 @@ class DebianBox(base.BaseDevice):
     shared_tftp_server = False
     wan_dhcp_server = True
     tftp_device = None
+    tftp_dir = '/tftpboot'
 
     iface_dut = "eth1"
     gw = None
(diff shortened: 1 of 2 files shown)


devices/debian.py
```diff
@@ -402,6 +402,41 @@ def start_sshd_server(self):
         self.sendline('/etc/init.d/ssh reload')
         self.expect(self.prompt)
 
+    def get_tftp_cfg_dest_path(self, board=None):
+        return '/tftpboot' # can this change?
```
mattsm: Not so far, but we can just make this an arg in the function below with the default being /tftpboot for now?

devices/debian.py
```diff
@@ -402,6 +402,41 @@ def start_sshd_server(self):
...
+
+    def get_conf_file_from_tftp(self, board, cfg):
+        """Retrieve a file from the tftp sever"""
+        if cfg is None:
+            return None
```
mattsm: Can we make the default action to download the CM and MTA cfg files? I assume this runs on the provisioner? We should add a function there to return the current configs (as read from DHCP config), so we can just use those filenames/paths

devices/debian.py
mgualco-contractor: ACK

devices/debian.py
mgualco-contractor: yes, we can, but wouldn't it be easier to get them directly from the device /var/tmp?

mattsm:
We talked more about this- any more questions?

mgualco-contractor:
they will come....

PR #256, opened 2019-02-27
tests: lib: docsis: add stub for cm_cfg and making changes

mattsm:

### DIFF
--- a/tests/lib/docsis.py
+++ b/tests/lib/docsis.py
@@ -9,6 +9,7 @@
 import os, config
 from common import cmd_exists
 import Tkinter
+import re
 
 class docsis:
     """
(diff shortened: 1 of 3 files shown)


tests/lib/docsis.py
```diff
@@ -84,3 +96,68 @@ def encode_cm():
...
+    legacy_search_path = None
+
+    def __init__(self, start=None):
+        '''Creates a default basic CM cfg file for modification'''
+
+        # TODO: we require loading a file for the moment
+        if start == None:
+            raise Exception("We don't know how to start from scratch yet")
+        else:
+            # TODO: filename should not matter
```
mgualco-contractor: Nit: maybe check that the file is a text file?

tests/lib/docsis.py
```diff
@@ -84,3 +96,68 @@ def encode_cm():
...
+            self.original_file = start
+            self.original_fname = os.path.split(start)[1]
```
mgualco-contractor: Nit: could this be `os.path.basename` ?

tests/lib/docsis.py
mattsm: If it does not exist it will throw an exception with useful info...

tests/lib/docsis.py
mattsm: Yes, it should... copying from input args... but for now I don't have a good test so I don't want to risk it in this commit.

PR #260, opened 2019-03-07
tests: lib: common: added function to enable/disable IPV6 in linux

premandfriends:
-The function will enable/disable IPv6 in linux clients based on value 1/0
   -1 to Disable, 0 to enable

tests: lib: docsis: added method to check CM is in briged mode

   -Under cm_cfg class added method "_is_bridged" to check whether CM is in bridge mode

Signed-off-by: Prem Kumar K N <prekumar.contractor@libertyglobal.com>

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -19,8 +19,6 @@ import re
 
 from termcolor import colored, cprint
 
-from lib.docsis import docsis, cm_cfg
-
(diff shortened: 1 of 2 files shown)


mattsm:
This looke perfect, except that it should have been two commits... can you split it up please?

PR #261, opened 2019-03-08
zephyr: zephyr_reporter: modify cycleName rule

lynnlincbn:
Base on commit REMOVED PRIVATE URL to modify cycleName rule

---
This change is [Reviewable](https://reviewable.io/reviews/lgirdk/boardfarm/261)

### DIFF
--- a/zephyr/zephyr_reporter.py
+++ b/zephyr/zephyr_reporter.py
@@ -100,7 +100,7 @@ def parse_zapi_config():
 
     return data
 
-def update_zephyr(test_cases_list):
+def update_zephyr(test_cases_list, result_data):
     args=parse_zapi_config()
     if len(args) == 0:
         print("Zephyr is not configured, skipping...")
@@ -117,10 +117,20 @@ def update_zephyr(test_cases_list):
                     options={'server': z["jira_url"]})
 
         proj = jira.project(z["project"])
-        verid = get_jira_release_id(z['release'], jira, proj)
-        cycleName = z["cycle"]
-        cycleName = cycleName + "_" + str((datetime.datetime.now()).strftime("%Y%m%d%H%M%S"))
-
+        date = str((datetime.datetime.now()).strftime("%Y%m%d"))
+        try:
+            match = re.search('SW_REV: \w+\-\w+\-((\d+\.?)+(-\w+))-SH', str(result_data))
+            version = match.group(1).replace(match.group(3),'.0-PRE-CH7465LG')
+            if date in match.group(3):
+                cycleName = match.group(1)+'-SH-DailyBuild'
+            else:
+                cycleName = match.group(1)
+        except:
+            version = z["release"]
+            cycleName = z["cycle"]
+        print("version = %s" %version)
+        print("cycleName = %s" %cycleName)
+        verid = get_jira_release_id(version, jira, proj)
 
         reporter = zapi.Zapi(project_id=proj.id,
                              version_id=verid,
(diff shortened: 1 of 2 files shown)


mattsm:
I see this, and understand but need to sit down and take a closer look and test.. will take time to review.

lynnlincbn:
Hi Matthew, did you have any update for your review? thanks.

zephyr/zephyr_reporter.py
```diff
@@ -119,7 +119,7 @@ def update_zephyr(test_cases_list):
         proj = jira.project(z["project"])
         verid = get_jira_release_id(z['release'], jira, proj)
         cycleName = z["cycle"]
-        cycleName = cycleName + "_" + str((datetime.datetime.now()).strftime("%Y%m%d%H%M%S"))
+        cycleName = cycleName.replace("date", str((datetime.datetime.now()).strftime("%Y%m%d")))
```
mattsm: In the future I think we just need to change the config file from JSON to a python file and we can encode whatever logic we want in python for each config we use...

For example

```
import datetime

config = {
	"report": "output.xml",
	"metafile": "zephyr/boardfarm_tc_meta_file.csv",
	"project": "PROJECT",
	"release": "RELEASE",
	"environment": "Test",
	"cycle": "CYCLE_%s_ABC" % datetime.datetime.now()).strftime("%Y%m%d%H%M%S"),
	"build": "DemoBuild",
	"jira_url": "JIRA_URL",
	"user": "USER",
	"passwd": "PASS",
	"updateautomationstatus": true,
	"status_codes": {
		"UNEXECUTED": -1,
		"PASS": 1,
		"FAIL": 2,
		"NOT TESTED": 7
	}
}
```

mattsm: If a change is required can you change the date replace from 'date' to '<DATE>'

zephyr/zephyr_reporter.py
```diff
@@ -136,46 +136,47 @@ def update_zephyr(test_cases_list):
...
+                     exec_id,
+                     log_data,
+                     status_code_dict=z['status_codes'])
+                else:
+                    ret = reporter.set_execution(result,
+                     exec_id,
+                     log_data)
+
+                if ret.status_code != requests.codes.ok:
+                    raise Exception("Error = %s, when trying to set execution status" % ret)
```
mattsm: Is this really needed when the other change merged? 

https://github.com/lgirdk/boardfarm/commit/36b8fac7872ea11d9b2495451f360c83b1327cfd

zephyr/zephyr_reporter.py
mattsm:
```
<DATE>
```

zephyr/zephyr_reporter.py
lynnlincbn: I don't know what other change merged? Can you tell more detail for this?
This commit is to fix if test_cases_list[i] ==[] will error on line 139

zephyr/zephyr_reporter.py
lynnlincbn: Not only date need to modify, also "release" and "cycle" need to change.
But for now I modify those manually.
Can you help me how can I pass value from boardfarm to zephyr_reporter?
I can snmpget fw version and modify to release and cycle format, but can't pass those parameters now.

mattsm: We discusses this privately. Please use the info_for_remote dict and self.logged in the tests..

zephyr/zephyr_reporter.py
mattsm: I think you just need to fix the for loop:

for test_case in test_cases_list:

should be 

for i in range(len(test_cases_list)):

mattsm:
Does not seem to be interest in fixing this, closing for now.

PR #264, opened 2019-03-08
Pr common function

ppalaniappan:


tests/lib/common.py
```diff
@@ -440,8 +440,17 @@ def snmp_mib_get(device, board, iface_ip, mib_name, index, timeout=10, retry=3):
...
     return snmp_out
+
+def hex2ipv6(self, hexstr):
+    """
+    Can parse strings in this form:
+    FE 80 00 00 00 00 00 00 3A 43 7D FF FE DC A6 C3
+    """
+    hexstr = hexstr.replace(' ', '').lower()
+    blocks = (''.join(block) for block in zip(*[iter(hexstr)]*4))
+    return ':'.join(str(block) for block in blocks)
```
mattsm: Can we validate this with ipaddress.IPv6Address?

Even if we just resturn a string for now? I'd like to start using ipaddress for all handling of IPs going forward. If this is a new function maybe we can just return:

return ipaddress.IPv6Address(':'.join(str(block) for block in blocks).encode('utf-8'))

mattsm:
Those lines are not on the first line, and not in the subject line... so it makes the commits look like

```
* 8e35826 (HEAD, lgi/pr/264) tests: lib: regexlib
* 43699d1 tests: lib: common
* fe7fa6d (gerritforge/master) tests: lib: installers: bug fix
* 09c049f devices: debian_isc: force wan, mta, erouter to correct network
* 8d1517a (lgi/pr/260) devices: debian: refactor ISC provisioner into separate class
```

The first line should be used to explain in less than 80 chars a summary of the patch.

PR #266, opened 2019-03-11
bft: add better error messages for failing to parse tests specified i…

mattsm:
…n -e


mattsm:
Yes, with and without BFT_DEBUG and with and without a test that would have a parsing issue.

PR #268, opened 2019-03-12
bft: fix update the results in Zephyr issue

lynnlincbn:
Move test_cases_list.append([]) to tc for loop to fix list index out of range, Unable to Update results in Zephyr issue.

### DIFF
--- a/bft
+++ b/bft
@@ -433,10 +433,9 @@ def main():
         result_data = json.load(open('./results/test_results.json'))
         length = len(result_data["test_results"])
         test_cases_list = []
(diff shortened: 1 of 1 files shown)


mattsm:
this just fixes an error message right? it's not doing anything else??

lynnlincbn:
Hi Matthew, this commit just fix test case append issue.
If test case more than 2 test cases will cause index out of range fail.

mattsm:
Oh yea, pretty clear. We did never test with more than one test ;-)

PR #270, opened 2019-03-13
tests: rootfs_boot: add wan container ipv6 default gateway

lynnlincbn:
Due to wan container ipv6 default gateway sometimes will disappear issue, add ipv6 default gateway on RootFSBootTest before run test case.

### DIFF
--- a/tests/rootfs_boot.py
+++ b/tests/rootfs_boot.py
@@ -146,6 +146,12 @@ class RootFSBootTest(linux_boot.LinuxBootTest):
             if wan:
                 wan.sendline('\nip -6 route del default')
                 wan.expect(prompt)
+                # Add eth1 ipv6 default gateway
+                default_gateway_ipv6 = self.config.board['common']['Default_gateway_ipv6']
+                wan.sendline('\nip -6 route add default via %s dev eth1' %default_gateway_ipv6)
+                wan.expect(prompt)
+                wan.sendline('\nip -6 route show')
+                wan.expect(prompt)
 
         # Give other daemons time to boot and settle
         if self.config.setup_device_networking:


tests/rootfs_boot.py
```diff
@@ -146,6 +146,12 @@ def boot(self, reflash=True):
             if wan:
                 wan.sendline('\nip -6 route del default')
                 wan.expect(prompt)
+                # Add eth1 ipv6 default gateway
+                default_gateway_ipv6 = self.config.board['common']['Default_gateway_ipv6']
```
mgualco-contractor: this failed on the unit test, maybe if 'common' exists add the route?

lynnlincbn:
This commit need to use below cbn.json to get the ipv6 default gateway parameter.
http://172.16.1.237:8080/c/boardfarm-cbn/+/2540

mattsm:
I'm going to be looking at lots of ipv6 fixes next week, I will keep this open to make sure it's covered as well.

But you can't assume this value is in a config ever, so it would need a sane default. That begin said isn't ipv6 supposed to handle this discovery?

tests/rootfs_boot.py
mattsm: Doesn't ipv6 routes get set by router advertisements? Why isn't your environment setup properly.

Can you review @ketantewari

ketantewari:
I guess it's the L3 switch configuration issue, which needs to be resolved with the network team. Coz the RA from the switch doesn't provide the default route option / prefix details / DRP.

mattsm:
> This commit need to use below cbn.json to get the ipv6 default gateway parameter.
> http://172.16.1.237:8080/c/boardfarm-cbn/+/2540

This commit is private to you, it should not be required for boardfarm to run standalone

lynnlincbn:
Will close this commit when manual team follow Ketan's suggestion to add route “2001:DB50::/32” in Switch.

PR #271, opened 2019-03-13
deploy-boardfarm-nodes: Ketan's changes

mgualco-contractor:
moved the default route from eth0 to table 102
on Ketan's behalf

### DIFF
--- a/deploy-boardfarm-nodes.sh
+++ b/deploy-boardfarm-nodes.sh
@@ -30,10 +30,11 @@ END
 
 local_route () {
 	# TODO: This is a problem if the router network matches the host network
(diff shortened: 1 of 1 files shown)


mattsm:
Will connections getting established in from ssh go via this route? E.g. can I still ssh to the container?

mattsm:
@ketantewari please provide answers...

ketantewari:
Yes, we tried it out it works

mattsm:
But then won't basic things fail? Like apt install? Is there a way to force a command to use a specific routing table in Linux?

mattsm:
Holing off on this, since we're discussing using net namespaces instead.

PR #273, opened 2019-03-14
tests: lib: installers: modified install_postfix function

premandfriends:
- latest installation of postfix in linux do not expect parameters
    - Modified the function to work both with/without parameters

Signed-off-by: prekumar.contractor<prekumar.contractor@libertyglobal.com>


mattsm:
We should consider locking down the version so this does not occur frequently...

PR #275, opened 2019-03-14
tests: rootfs_boot: fix logic for checking IP address

mattsm:
We can't run the assert if we don't have an IP, we should just loop
again


mattsm:
please review very carefully- maybe even test

mgualco-contractor:
added a comment via gerrithub

mattsm:
Merged via gerrithub

PR #282, opened 2019-03-26
tests: lib: wifi: Added get functions in wifi_stub

sravichandran-contractor:


tests/lib/wifi.py
```diff
@@ -197,3 +197,21 @@ def set_bandwidth(self, *args, **kwargs):
         raise Exception("Not implemented!")
     def set_channel_number(self, *args, **kwargs):
         raise Exception("Not implemented!")
+    def get_wifi_enable(self, *args, **kwargs):
```
mattsm: enableD ?

sravichandran-contractor: Yes, changed it to enabled

PR #283, opened 2019-03-26
devices: casa_cmts: added function to get erouter ipv4 and ipv6 from cmts

ppalaniappan:
Signed-off-by: Panimalar Palaniappan <ppalanaiappan.contractor@libertyglobal.com>

### DIFF
--- a/devices/casa_cmts.py
+++ b/devices/casa_cmts.py
@@ -11,6 +11,7 @@ import sys
 import base
 import re
 import connection_decider
+from lib.regexlib import ValidIpv6AddressRegex, ValidIpv4AddressRegex
 
 class CasaCMTS(base.BaseDevice):
     '''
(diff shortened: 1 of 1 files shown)


devices/casa_cmts.py
```diff
@@ -446,6 +447,13 @@ def check_docsis_mac_ip_provisioning_mode(self, index):
         if self.match != None:
             return result
 
+    def check_cpe_mode(self, mac):
```
mattsm: Is this really checking a mode or getting ipv4 and ipv6 addresses?

mattsm: Can we just make two new functions for this? get_cpe_ipv4_addr? get_cpe_ipv6_addr?

devices/casa_cmts.py
```diff
@@ -446,6 +447,26 @@ def check_docsis_mac_ip_provisioning_mode(self, index):
...
+
+    def get_ertr_ipv6(self, mac):
+        self.sendline("show cable modem %s cpe" % mac)
+        self.expect(self.prompt)
+        ertr_ipv6 = re.search(ValidIpv6AddressRegex ,self.before)
+        if ertr_ipv6:
+            ipv6 = ertr_ipv6.group()
+            return ipv6
+        else:
+            return None
```
mattsm: I don't want to keep beating a dead horse, but should these really return None or raise an Exception?

I will merge as-is for now, but please make sure this is thought out properly.

PR #284, opened 2019-03-26
devices: debian: install xxd in device (copy_file_to_server related)

mgualco-contractor:
When copying a file to a server the device did not have xxd installed
and the copy failed without an error message.
This makes sure xxd is installed.


mattsm:
Can we move this to the install_pkgs function? It's a mess for debian for now but I want to keep it in one place..

mgualco-contractor:
yes, i was being lazy

mattsm:
> yes, i was being lazy

I don't see that as being harder than this ;p

PR #286, opened 2019-03-28
Pr cmts prompt regex issue

mgualco-contractor:
this should avoid part of the telnet login message ending up in the casa cmts prompt list.
tested on 3 different versions of casa (3000, 3200,10G)

---
This change is [Reviewable](https://reviewable.io/reviews/lgirdk/boardfarm/286)

### DIFF
--- a/devices/casa_cmts.py
+++ b/devices/casa_cmts.py
@@ -41,12 +41,12 @@ class CasaCMTS(base.BaseDevice):
 
     def connect(self):
         try:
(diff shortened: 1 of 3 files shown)


mattsm:
I'm taking the first two commits for the telnet regex and then this

https://review.gerrithub.io/c/lgirdk/boardfarm/+/449492/2/devices/casa_cmts.py

instead of your third one for now.. we can revisit if we see these issues.

mgualco-contractor:
ACK

PR #307, opened 2019-04-30
devices: debian_wifi: Enhanced the wifi functions to execute in sudo as well

sravichandran-contractor:
This change is [Reviewable](https://reviewable.io/reviews/lgirdk/boardfarm/307)

### DIFF
--- a/devices/debian_wifi.py
+++ b/devices/debian_wifi.py
@@ -1,88 +1,111 @@
 import re
 import debian
 import pexpect
+from countrycode import countrycode
+from lib.wifi import wifi_client_stub
 
-class DebianWifi(debian.DebianBox):
+class DebianWifi(debian.DebianBox, wifi_client_stub):
     '''Extension of Debian class with wifi functions'''
 
     model = ('debianwifi')
+    iface_wlan = "wlan0"
+    iface_wlan1 = "wlan1"
 
-    iface_wlan = "wlan1"
+    def disable_and_enable_wifi(self, iface):
+       self.set_link_state(iface, "down")
+       self.set_link_state(iface, "up")
 
-    def scan(self):
+    def disable_wifi(self, iface):
+       self.set_link_state(iface, "down")
+
+    def enable_wifi(self, iface):
+       self.set_link_state(iface, "up")
+
+    def wifi_scan(self, iface):
         from tests.lib.installers import install_iw
         install_iw(self)
 
-        self.sendline('iw %s scan | grep SSID:' % self.iface_wlan)
+        self.sudo_sendline('iw %s scan | grep SSID:' % iface)
         self.expect(self.prompt)
-
         return self.before
 
-    def wpa_connect(self, ssid_name, password):
-         '''Generate WPA supplicant file and execute it'''
-         self.sendline("rm /etc/"+ssid_name+".conf")
-         self.expect(self.prompt)
-         self.sendline("wpa_passphrase "+ssid_name+ " >> /etc/"+ssid_name+".conf")
-         self.expect("")
-         self.sendline(password)
-         self.expect(self.prompt,timeout=20)
-         self.sendline("cat /etc/"+ssid_name+".conf")
-         self.expect(self.prompt)
-         self.sendline("wpa_supplicant  -B -Dnl80211 -i"+self.iface_wlan+ " -c/etc/"+ssid_name+".conf")
-         self.expect(self.prompt)
-         wpa_start = None
-         wpa_start = self.before
-         match = re.search('Successfully initialized wpa_supplicant',wpa_start)
-         if match:
-            return match.group(0)
-         else:
-            return None
+    def wifi_check_ssid(self, iface, ssid_name):
+        from tests.lib.installers import install_iw
+        install_iw(self)
 
-    def wlan_connectivity(self):
-        '''Connection state verify'''
-        self.sendline("iw %s link" % self.iface_wlan)
+        self.sudo_sendline('iw %s scan | grep "SSID: %s"' % (iface, ssid_name))
         self.expect(self.prompt)
-        conn_state = self.before
-        match = re.search('Connected',conn_state)
+        match = re.search("%s\"\s+.*(%s)"%(ssid_name, ssid_name), self.before)
+        if match:
+            return True
+        else:
+            return False
+
+    def wifi_connect(self, iface, ssid_name, password=None, security_mode=None):
+        if password == None:
+            self.sudo_sendline("iwconfig %s essid %s" % (iface,ssid_name))
+        else:
+            '''Generate WPA supplicant file and execute it'''
+            self.sudo_sendline("rm "+ssid_name+".conf")
+            self.expect(self.prompt)
+            self.sudo_sendline("wpa_passphrase "+ssid_name+" "+password+" >> "+ssid_name+".conf")
+            self.expect(self.prompt)
+            self.sendline("cat "+ssid_name+".conf")
+            self.expect(self.prompt)
+            self.sudo_sendline("wpa_supplicant  -B -Dnl80211 -i"+iface+ " -c"+ssid_name+".conf")
+            self.expect(self.prompt)
+            match = re.search('Successfully initialized wpa_supplicant', self.before)
+            if match:
+                return True
+            else:
+                return False
+
+    def wifi_connectivity_verify(self, iface):
+        '''Connection state verify'''
+        self.sendline("iw %s link" % iface)
+        self.expect(self.prompt)
+        match = re.search('Connected', self.before)
+        if match:
+            return True
+        else:
+            return False
+
+    def disconnect_wpa(self):
+        self.sudo_sendline("killall wpa_supplicant")
+        self.expect(self.prompt)
+
+    def wlan_ssid_disconnect(self, iface):
+        output = self.sudo_sendline("iw dev %s disconnect" % iface)
+        self.expect(self.prompt)
+
+    def wifi_disconnect(self, iface):
+        self.disconnect_wpa()
+        self.wlan_ssid_disconnect(iface)
+
+    def wifi_change_region(self, country):
+        country = countrycode(country, origin='country_name', target='iso2c')
+        self.sudo_sendline("iw reg set %s"%(country))
+        self.expect(self.prompt)
+        self.sendline("iw reg get")
+        self.expect(self.prompt)
+        match = re.search(country, self.before)
         if match:
             return match.group(0)
         else:
             return None
 
-    def disconnect_wpa(self):
-        self.sendline("killall wpa_supplicant")
-        self.expect(self.prompt)
-
-    def wlan_ssid_connect(self, ssid_name):
-        self.sendline("iwconfig %s essid %s" % (self.iface_wlan,ssid_name))
-        self.expect(self.prompt)
-
-    def wlan_ssid_disconnect(self):
-        self.sendline("iw dev %s disconnect" % self.iface_wlan)
-        self.expect(self.prompt)
-
-    def start_lan_client(self, wan_gw=None):
-        self.iface_dut = self.iface_wlan
+    def start_lan_client(self, iface):
+        self.iface_dut = iface
         super(DebianWifi, self).start_lan_client()
 
-    def linux_wifi_client(self, board, ssid_name, password=None):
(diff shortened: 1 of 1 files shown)


devices/debian_wifi.py
```diff
@@ -7,29 +7,38 @@ class DebianWifi(debian.DebianBox):
...
+        if match:
+            self.sendline('sudo iw %s scan | grep SSID:' % iface_wlan)
+            idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+            if idx == 0:
+                self.sendline(login_pwd)
+                self.expect(self.prompt)
         return self.before
 
-    def wpa_connect(self, ssid_name, password):
+    def wpa_connect(self, ssid_name, password, iface_wlan="wlan1", login_pwd=None):
```
mattsm: What is the iface for? Don't we have self.iface?

Is this for two wireless interfaces?

devices/debian_wifi.py
```diff
@@ -7,29 +7,38 @@ class DebianWifi(debian.DebianBox):
...
          self.expect(self.prompt)
-         self.sendline("wpa_passphrase "+ssid_name+ " >> /etc/"+ssid_name+".conf")
-         self.expect("")
-         self.sendline(password)
-         self.expect(self.prompt,timeout=20)
-         self.sendline("cat /etc/"+ssid_name+".conf")
+         self.sendline("cat "+ssid_name+".conf")
          self.expect(self.prompt)
-         self.sendline("wpa_supplicant  -B -Dnl80211 -i"+self.iface_wlan+ " -c/etc/"+ssid_name+".conf")
+         self.sendline("wpa_supplicant  -B -Dnl80211 -i"+iface_wlan+ " -c"+ssid_name+".conf")
```
mattsm: Yea, so why are we switching away? If there is a new one, we need to handle two radios somehow...

devices/debian_wifi.py
```diff
@@ -7,29 +7,38 @@ class DebianWifi(debian.DebianBox):
...
          self.expect(self.prompt)
+         match = re.search("Failed to initialize driver interface",self.before)
+         if match:
+             self.sendline("sudo wpa_supplicant  -B -Dnl80211 -i"+iface_wlan+ " -c"+ssid_name+".conf")
+             idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+             if idx == 0:
+                 self.sendline(login_pwd)
+                 self.expect(self.prompt)
```
mattsm: See sudo_sendline()

devices/debian_wifi.py
```diff
@@ -38,51 +47,77 @@ def wpa_connect(self, ssid_name, password):
...
+                self.sendline(login_pwd)
+                wlan_connect = self.expect('Connected', timeout=5)
+                self.expect(self.prompt)
+        if wlan_connect == 0:
+            return "Connected"
         else:
             return None
 
-    def disconnect_wpa(self):
+    def disconnect_wpa(self, login_pwd=None):
```
mattsm: Why the added arg? Do we plan to use it?

devices/debian_wifi.py
```diff
@@ -38,51 +47,77 @@ def wpa_connect(self, ssid_name, password):
...
-
-    def wlan_ssid_connect(self, ssid_name):
-        self.sendline("iwconfig %s essid %s" % (self.iface_wlan,ssid_name))
+        match = re.search("Operation not permitted", self.before)
+        if match:
+            self.sendline("sudo killall wpa_supplicant")
+            idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+            if idx == 0:
+                self.sendline(login_pwd)
+                self.expect(self.prompt)
```
mattsm: sudo_sendline()

devices/debian_wifi.py
```diff
@@ -38,51 +47,77 @@ def wpa_connect(self, ssid_name, password):
...
+
+    def wlan_ssid_connect(self, ssid_name, iface_wlan="wlan1", login_pwd=None):
+        self.sendline("iwconfig %s essid %s" % (iface_wlan,ssid_name))
         self.expect(self.prompt)
```
mattsm: We don't use login_pwd? What is this value for?

devices/debian_wifi.py
```diff
@@ -38,51 +47,77 @@ def wpa_connect(self, ssid_name, password):
...
-
-    def wlan_ssid_disconnect(self):
-        self.sendline("iw dev %s disconnect" % self.iface_wlan)
+        match = re.search("Operation not permitted", self.before)
+        if match:
+            self.sendline("sudo iwconfig %s essid %s" % (iface_wlan,ssid_name))
+            idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+            if idx == 0:
+                self.sendline(login_pwd)
+                self.expect(self.prompt)
```
mattsm: sudo_sendline()

devices/debian_wifi.py
```diff
@@ -38,51 +47,77 @@ def wpa_connect(self, ssid_name, password):
...
-
-    def start_lan_client(self, wan_gw=None):
-        self.iface_dut = self.iface_wlan
+        match = re.search("Operation not permitted", self.before)
+        if match:
+            self.sendline("sudo iw dev %s disconnect" % iface_wlan)
+            idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+            if idx == 0:
+                self.sendline(login_pwd)
+                self.expect(self.prompt)
```
mattsm: sudo_sendline()

devices/debian_wifi.py
sravichandran-contractor: Ack

devices/debian_wifi.py
sravichandran-contractor: Replaced to take it from cbn.json

devices/debian_wifi.py
sravichandran-contractor: Yes we have 2 interfaces. Added them in Debianwifi class

devices/debian_wifi.py
```diff
@@ -3,10 +3,9 @@
 import pexpect
 from countrycode import countrycode
 from lib.wifi import wifi_client_stub
-import pdb
 
 class DebianWifi(debian.DebianBox, wifi_client_stub):
-    '''Extension of Debian class with wifi functions'''
+    ''' Extension of Debian class with wifi functions '''
```
mattsm: This does not seem required.

devices/debian_wifi.py
```diff
@@ -3,10 +3,9 @@
 import pexpect
 from countrycode import countrycode
 from lib.wifi import wifi_client_stub
-import pdb
```
mattsm: Why are you doing this? This commit is a copy of the previous commit message yet it's only contents are this?

devices/debian_wifi.py
```diff
@@ -1,88 +1,126 @@
...
-class DebianWifi(debian.DebianBox):
+class DebianWifi(debian.DebianBox, wifi_client_stub):
     '''Extension of Debian class with wifi functions'''
 
     model = ('debianwifi')
+    wifi_interface = "wlan0"
+    wifi_interface1 = "wlan1"
 
-    iface_wlan = "wlan1"
+    wifi_client_control = None
```
mattsm: We have stuck with iface_ all around why change it now?

devices/debian_wifi.py
```diff
@@ -1,88 +1,126 @@
...
 
-    def scan(self):
+    def disable_and_enable_wifi(self, iface):
```
mattsm: On the stub commit I suggest breaking these into enable/disable functions not combining both. They will be generically more useful.

devices/debian_wifi.py
```diff
@@ -1,88 +1,126 @@
...
+
+    def wifi_scan(self, iface, ssid_name=None):
         from tests.lib.installers import install_iw
         install_iw(self)
 
-        self.sendline('iw %s scan | grep SSID:' % self.iface_wlan)
-        self.expect(self.prompt)
-
+        self.sudo_sendline('iw %s scan | grep SSID:' % iface)
+        idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
```
mattsm: I think you missed the entire point of sudo_sendline(). It's supposed to handle all aspects of the password prompt.

mattsm:
Also, this is now two commits... with the second commit not making any sense. Please squash to a single commit.

devices/debian_wifi.py
sravichandran-contractor: This is a wifi control to set wifi or wlan. I haven't removed the iface_ which we have been using

devices/debian_wifi.py
sravichandran-contractor: Ack

devices/debian_wifi.py
sravichandran-contractor: Will change this

sravichandran-contractor:
> Also, this is now two commits... with the second commit not making any sense. Please squash to a single commit.

Not sure how I missed. Have squashed it

devices/debian_wifi.py
```diff
@@ -1,88 +1,109 @@
...
     '''Extension of Debian class with wifi functions'''
 
     model = ('debianwifi')
+    iface_wlan = "wlan0"
+    iface_wlan1 = "wlan1"
 
-    iface_wlan = "wlan1"
+    def disable_and_enable_wifi(self, iface):
+       self.set_link_state(iface,"down")
+       self.set_link_state(iface,"up")
```
mattsm: Can you fix the space after the , ?

I would just all disable_wifi() and enable_wifi() incase those functions ever do something else.

devices/debian_wifi.py
```diff
@@ -1,88 +1,109 @@
...
 
-    def scan(self):
+    def disable_wifi(self, iface):
+       self.set_link_state(iface,"down")
+
+    def enable_wifi(self, iface):
+       self.set_link_state(iface,"up")
```
mattsm: no spacing after commas

devices/debian_wifi.py
```diff
@@ -1,88 +1,109 @@
...
+
+    def wifi_scan(self, iface):
         from tests.lib.installers import install_iw
         install_iw(self)
 
-        self.sendline('iw %s scan | grep SSID:' % self.iface_wlan)
-        self.expect(self.prompt)
+        output = self.sudo_sendline('iw %s scan | grep SSID:' % iface)
```
mattsm: If you want the output, you should do an expect.

sendline() never returns matched chars, just the number of bytes sent.

devices/debian_wifi.py
```diff
@@ -1,88 +1,109 @@
...
 
-    def wlan_connectivity(self):
+        output = self.sudo_sendline('iw %s scan | grep "SSID: %s"' % (iface,ssid_name))
+        match = re.search("%s\"\s+.*(%s)"%(ssid_name,ssid_name),output)
+        if match:
+            return True
+        else:
+            return False
+
+    def wifi_connect(self, iface, ssid_name, password=None, security_mode=None):
```
mattsm: Why do you even need the password arg here? It should all be handled properly by sudo_sendline()

devices/debian_wifi.py
```diff
@@ -1,88 +1,109 @@
...
-        assert conn_wlan!=None,'Connection establishment in WIFI'
-
-        self.start_lan_client()
+        self.disable_and_enable_wifi(iface)
+        time.sleep(10)
+        output = self.wifi_check_ssid(iface,ssid_name)
+        assert output==True,'SSID value check in WLAN container'
+
+        conn_wifi = self.wifi_connect(iface,ssid_name,password)
+        time.sleep(20)
```
mattsm: nit: can we poll instead of just straight up sleeping? also self.expect(pexpect.TIMEOUT, timeout=20) is better than just sleeping

devices/debian_wifi.py
sravichandran-contractor: Ack

devices/debian_wifi.py
sravichandran-contractor: Ack

devices/debian_wifi.py
sravichandran-contractor: This is the wifi password

mattsm: oh you are right, I was still stuck on sudo password... this is OK then.

PR #308, opened 2019-04-30
devices: debian: Enhanced link_up() and set_link_state()

sravichandran-contractor:
Modified the functions to perform in sudo as well

---
This change is [Reviewable](https://reviewable.io/reviews/lgirdk/boardfarm/308)

### DIFF
--- a/devices/debian.py
+++ b/devices/debian.py
@@ -205,10 +205,19 @@ class DebianBox(base.BaseDevice):
         cc.expect(pexpect.EOF, timeout=120)
         print("cleanup_cmd done.")
 
-    def sudo_sendline(self, s):
+    def sudo_sendline(self, cmd):
         if self.username != "root":
-            s = "sudo " + s
-        return super(DebianBox, self).sendline(s)
+            self.sendline("sudo true")
+            if 0 == self.expect(["password for .*:"] + self.prompt):
+                 will_prompt_for_password = True
+            else:
+                 will_prompt_for_password = False
+
+            cmd = "sudo " + cmd
+            if will_prompt_for_password:
+                self.sendline(self.password)
+                self.expect(self.prompt)
+        super(DebianBox, self).sendline(cmd)
 
     def reset(self):
         self.sendline('reboot')
@@ -752,7 +761,7 @@ EOFEOFEOFEOF''' % (dst, bin_file))
         self.sendline("ip link show %s" % interface)
         self.expect(self.prompt)
         link_state = self.before
-        match = re.search('NO-CARRIER,BROADCAST,MULTICAST,UP',link_state)
+        match = re.search('BROADCAST,MULTICAST,UP',link_state)
         if match:
             return match.group(0)
         else:
@@ -760,7 +769,7 @@ EOFEOFEOFEOF''' % (dst, bin_file))
 
     def set_link_state(self, interface, state):
         '''Setting the interface status'''
-        self.sendline("ip link set %s %s" % (interface,state))
+        self.sudo_sendline("ip link set %s %s" % (interface,state))
         self.expect(self.prompt)
 
 if __name__ == '__main__':


devices/debian.py
```diff
@@ -752,16 +752,23 @@ def link_up(self, interface):
...
         '''Setting the interface status'''
         self.sendline("ip link set %s %s" % (interface,state))
         self.expect(self.prompt)
+        match = re.search("Operation not permitted", self.before)
+        if match:
+            self.sendline("sudo ip link set %s %s" % (interface,state))
+            idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+            if idx == 0:
+                self.sendline(login_pwd)
+                self.expect(self.prompt)
```
mattsm: Check out sudo_sendline() in this class.

devices/debian.py
```diff
@@ -752,16 +752,23 @@ def link_up(self, interface):
...
         link_state = self.before
-        match = re.search('NO-CARRIER,BROADCAST,MULTICAST,UP',link_state)
+        match = re.search('BROADCAST,MULTICAST,UP',link_state)
         if match:
             return match.group(0)
         else:
             return None
 
-    def set_link_state(self, interface, state):
+    def set_link_state(self, interface, state, login_pwd=None):
```
mattsm: Also, why do we pass login_pwd here?

devices/debian.py
```diff
@@ -752,16 +752,19 @@ def link_up(self, interface):
...
             return match.group(0)
         else:
             return None
 
     def set_link_state(self, interface, state):
         '''Setting the interface status'''
-        self.sendline("ip link set %s %s" % (interface,state))
-        self.expect(self.prompt)
+        self.sudo_sendline("ip link set %s %s" % (interface,state))
+        idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
```
mattsm: This logic should go INSIDE sudo_sendline() not be repeated everywhere...

sravichandran-contractor: Ack

devices/debian.py
```diff
@@ -208,7 +208,12 @@ def run_cleanup_cmd(self):
     def sudo_sendline(self, s):
         if self.username != "root":
             s = "sudo " + s
-        return super(DebianBox, self).sendline(s)
+        super(DebianBox, self).sendline(s)
+        idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
+        if idx == 0:
+            self.sendline(self.password)
+            self.expect(self.prompt)
+        return self.before
```
mattsm: Was there a particular reason to return self.before vs. the "number of bytes written" that sendline in the pexpect library would have sent?

I think you should save the return value from 'super(DebianBox, self).sendline(s)' and probably return that instead.

sravichandran-contractor: The self.before returns the complete output hence I can use it for verification in the places I call this function, whereas if I store the  return value of super(DebianBox, self).sendline(s) and return it, it just gives me bytes.
For Eg: sudo iw wlan1 scan | grep "SSID: " would return the list of SSID which I can fetch from self.before

mattsm: As I commented in your other change, I think we need to just call expect after this function to exact the desired text. I think that means you need to remove the self.expect(self.prompt) so you don't eat that.

sudo_sendline() should not expect anything after the password related messages.

devices/debian.py
```diff
@@ -208,7 +208,12 @@ def run_cleanup_cmd(self):
     def sudo_sendline(self, s):
         if self.username != "root":
             s = "sudo " + s
-        return super(DebianBox, self).sendline(s)
+        super(DebianBox, self).sendline(s)
+        idx = self.expect(["password for .*:"] + self.prompt,timeout=20)
```
mattsm: So this is super tricky actually. You will need to avoid eating the output somehow...

I think you will have to run a test command to see if sudo prompts for a password
```
self.sendline(sudo true)
if 0 == expect(password for [^:]*:, prompt):
    will_prompt_for_password = True
else
    will_prompt_for_password = False

self.sendline(sudo + cmd)
if will_prompt_for_password:
    self.expect("password for .*:")
    self.sendline(self.password)
```

sravichandran-contractor: Yes it expects for password at times. I have tried it, that is the reason I added the password expect prompt here

mattsm:
Perfect! Thanks.

PR #315, opened 2019-05-06
tests: lib: wifi: Added wifi_client_stub()

sravichandran-contractor:
This change is [Reviewable](https://reviewable.io/reviews/lgirdk/boardfarm/315)

### DIFF
--- a/tests/lib/wifi.py
+++ b/tests/lib/wifi.py
@@ -219,3 +219,51 @@ class wifi_stub():
         pass
     def cleanup(self):
         pass
+
+class wifi_client_stub():
+    def enable_wifi(self, iface):
+        '''Function to make the wifi interface UP
+           Takes interface as an argument'''
+        raise Exception("Not implemented!")
+    def disable_wifi(self, iface):
+        '''Function to make the wifi interface DOWN
+           Takes interface as an argument'''
+        raise Exception("Not implemented!")
+    def disable_and_enable_wifi(self):
+        '''Function to make the wifi interface DOWN and UP
+           Takes interface as an argument'''
+        raise Exception("Not implemented!")
+    def wifi_scan(self, iface):
+        '''Function that scans for SSIDs on a particular radio, and return a list of SSID
+           Takes interface to be scanned as an argument'''
+        raise Exception("Not implemented!")
+        # this code does not execute, but rather serves as an example for the API
+        return "SSID: <ssid_name1> \
+                SSID: <ssid_name2>.."
+    def wifi_check_ssid(self, iface, ssid_name):
+        '''Function that scans for a particular SSID
+           Takes interface and ssid to be scanned as an argument'''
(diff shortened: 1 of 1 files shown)


tests/lib/wifi.py
```diff
@@ -219,3 +219,19 @@ def prepare(self):
         pass
     def cleanup(self):
         pass
+
+class wifi_client_stub():
+    def disable_and_enable_wifi(self, *args, **kwargs):
```
mattsm: Can we make this enable_wifi() and disable_wifi() - it should be more  generically useful.

tests/lib/wifi.py
```diff
@@ -219,3 +219,19 @@ def prepare(self):
...
+        raise Exception("Not implemented!")
+    def wifi_connect(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def wifi_connectivity_verify(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def wifi_disconnect(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def wifi_change_region(self, *args, **kwargs):
+        raise Exception("Not implemented!")
+    def wifi_ssid_disconnect(self, *args, **kwargs):
```
mattsm: Why args and kwargs? These need to be fixed. Also it might be nice the leave an example return after the exception to standardize the format of the return values

E.g.  wifi_scan returns a list of dicts with the SSID name and other info retrived from the scan.

sravichandran-contractor: I followed the stub format which we have already used in wifi_stub(). Will work on this.
Am still not getting the example for return to be set after exception. Please help me on this

mattsm: I think for wifi_change_region we should suggest default arguments... maybe nothing on the latter one. Not sure why we needed these for the other stub.

sravichandran-contractor: I have removed the args and kwargs, but still should I be adding the arguments used in the function?

Also I couldn't get the "example return after exception" which you have mentioned. Could you provide an example for the same and give me more clarification on this?

mattsm: yea, so this is a stub, so it should be the basis for providing a reasonable first step to implementing the feature. so know what return values are expected to be would be nice.
```
def wifi_scan(self):
     '''Function that scans for SSIDs on a particular radio, takes no arguments and
        returns a dict of ssids with a dict of attributes for an ssid (e.g. channel, snr, etc)'''
     raise Exception(...)

     # this code does not execute, but rather serves as an example for the API
     return {ssid1: { "attr1": "val1", "attr2": "val2" },
                 ssid2: {  "attr1": "val1", "attr2": "val2" } }
```

sravichandran-contractor: Thanks a lot!! Got the point.. Will incorporate the same

PR #360, opened 2025-09-22
fix(devices.py): add specification for boardfarm_server_boot_async()

karrukola:
without this change, when an environment without server devices, the framework tries to call it, but the method is not defined, leading to a stack trace...


karrukola:
Another option is, of course, to avoid calling the async version of this method.
Let me know.

PR #361, opened 2025-09-22
refactor: implement src layout

karrukola:
enabler for further improvements, i.e. adoption different build systems.


src/boardfarm3/boardfarm3/__init__.py
ketantewari, line 1: The path of boardfarm3 package should have been 
src/boardfarm3/__init__.py

But currently it's:
src/boardfarm3/boardfarm3/__init__.py

Is this intentional?
Coz package installation fails for me.

ketantewari: Also, I don't see the pyproject.toml and noxfile changes that were previously available on the boardfarm3 branch's PR.

ketantewari (changes requested):

